PR #3, opened 2023-09-06
Feat: Add inputs pagination

brunomenezes:
#### Summary
Code changes to add pagination to the list of inputs. 

Changes:
* Added pagination to the list of inputs leveraging the URL query params.
* Added the feature to display from 10 to 30 items. Handle edge cases and update pagination buttons.


apps/web/src/app/page.tsx
```diff
@@ -39,47 +48,122 @@ const Summary = ({ inputs, applications }: SummaryProps) => {
...
-            <Summary
-                inputs={stats?.inputsConnection.totalCount ?? 0}
-                applications={stats?.applicationsConnection.totalCount ?? 0}
-            />
+            <Summary />
 
             <Divider
-                labelPosition="left"
+                ref={targetRef}
+                labelPosition="center"
```
tuler, line 92: I'd rather leave left centered. In general. I think it has more balance.

apps/web/src/app/page.tsx
```diff
@@ -39,47 +48,122 @@ const Summary = ({ inputs, applications }: SummaryProps) => {
 };
 
 const Explorer: FC = (props) => {
-    const [{ data: stats }] = useStatsQuery();
-    const [{ data }] = useInputsQuery();
+    const [{ limit, page }, updateParams] = usePaginationParams();
```
tuler, line 51: Can we use https://v7.mantine.dev/hooks/use-pagination

apps/web/src/app/page.tsx
```diff
@@ -1,27 +1,36 @@
...
+    InputOrderByInput,
+    useInputsQuery,
+    useStatsQuery,
+} from "../graphql/index";
+import { limitBounds, usePaginationParams } from "../hooks/usePaginationParams";
 
-const Summary = ({ inputs, applications }: SummaryProps) => {
+const Summary = () => {
+    const [{ data: stats }] = useStatsQuery();
+    const inputs = stats?.inputsConnection.totalCount ?? 0;
```
tuler, line 32: Missing a `?` here after `inputsConnection`

apps/web/src/app/page.tsx
```diff
@@ -1,27 +1,36 @@
...
+    const applications = stats?.applicationsConnection.totalCount ?? 0;
```
tuler, line 33: Missing a `?` here after `applicationsConnection`

apps/web/src/app/page.tsx
```diff
@@ -39,47 +48,122 @@ const Summary = ({ inputs, applications }: SummaryProps) => {
...
         <Stack>
             <Breadcrumbs>
                 <Anchor>Home</Anchor>
             </Breadcrumbs>
 
-            <Summary
-                inputs={stats?.inputsConnection.totalCount ?? 0}
-                applications={stats?.applicationsConnection.totalCount ?? 0}
-            />
+            <Summary />
```
tuler, line 88: I'd rather leave the state at the page, because then we can turn the `Summary` into a component at the UI library/storybook.

apps/web/src/app/page.tsx
```diff
@@ -39,47 +48,122 @@ const Summary = ({ inputs, applications }: SummaryProps) => {
...
-                    </Table.Tr>
-                </Table.Thead>
-                <Table.Tbody>
-                    {data?.inputs.map((input) => (
-                        <InputRow key={input.id} input={input} />
-                    ))}
-                </Table.Tbody>
-            </Table>
+
+            <Card shadow="md" withBorder px="xl" mt="md">
```
tuler, line 101: Remove all props from the `Card`. Leave at basics.

apps/web/src/providers/graphqlProvider.tsx
```diff
@@ -1,14 +1,20 @@
...
 import { FC } from "react";
 import { Client, Provider, cacheExchange, fetchExchange } from "urql";
-
 export type GraphQLProviderProps = {
     children?: React.ReactNode;
 };
 
+const GRAPHQL_URL = process.env.NEXT_PUBLIC_BE_GRAPHQL_URL ?? "";
+
+const url = not(isEmpty(GRAPHQL_URL))
```
tuler, line 10: Do we need this `ramda` library?
Isn't it just:
```
const url = process.env.NEXT_PUBLIC_BE_GRAPHQL_URL ?? "https://squid.subsquid.io/rollups-sepolia/v/v5/graphql";
```

apps/web/src/app/page.tsx
brunomenezes: They are not the same. This hook is about URL query params for a limit of items displayed and what is the page number. Maybe the name should be changed, but that sounds better than `useURLQueryParams` even more because I using the query param for a specific reason and values are not opened.

apps/web/src/app/page.tsx
brunomenezes: I could add. The one that does not exist is the stats/data Once you have a response back you have `inputsConnection` which is an object and totalCount is `Int!

apps/web/src/app/page.tsx
brunomenezes: I could add. The one that does not exist is the stats/data. Once you have a response back, you have `applicationConnection`, which is an object, and totalCount is `Int!

apps/web/src/app/page.tsx
brunomenezes: hummm. I thought the `Summary Card` was exported to the UI for this purpose, not this `Summary` component where we define the Grid with one or two columns.

apps/web/src/providers/graphqlProvider.tsx
brunomenezes: That is like a utility library like Lodash. I want to use all the other functions it offers, like `pathOr` that I am using on the page.

tuler: Ok, but is it useful in this call?

apps/web/src/app/page.tsx
tuler: Both can be there.

apps/web/src/app/page.tsx
tuler: Ok, but it's breaking on loading.

apps/web/src/app/page.tsx
tuler: Ok, but it's breaking on loading.

apps/web/src/app/page.tsx
tuler: But can we use the mantine hook?

brunomenezes: Yes, you can. another PR to research how to use it properly? None of the examples show both together is just one or another.

apps/web/src/app/page.tsx
brunomenezes: That is odd @tuler, but here [6f19eb1](https://github.com/cartesi/rollups-explorer/pull/3/commits/6f19eb1e7feb29e62633cab201416027169db66b) is the extra optional chaining check.

apps/web/src/app/page.tsx
brunomenezes: Ok, 20e86313977cf599ed0787c12594fbc9be4470ef. I don't think is so reusable, but the way it was before the storybook rewrite, where it gets a defined list of `SummaryItem` would be better if adjustments for number of columns / breakpoints and etc.

apps/web/src/app/page.tsx
brunomenezes: b8191427022167275905644c64f211077875a090

apps/web/src/providers/graphqlProvider.tsx
brunomenezes: a0723681cb0371821fa41a3747b64f4571269784

PR #7, opened 2023-09-13
Feature: Add programmatic codegen

brunomenezes:
### Summary
Programmatic codegen for graphQL using environment var to define schema location (accepts the usual local file or URL).

Changes: 
- Replace codegen.yaml with codegen.ts  using the programmatic API. 
- Update the backend env var from `NEXT_PUBLIC_BE_GRAPHQL_URL` to `NEXT_PUBLIC_EXPLORER_API_URL`.
- Also replace in the graphQL provider and also check if not defined, an error will be thrown.
- Add dotenv-cli for development convenience that works alongside a .env defined by the developer.

Deployment changes:
- Vercel is already updated with the new backend environment variable name.
- Also include the `NEXT_PUBLIC_ENABLE_TESTNETS` in the config for the Vercel Sepolia project.


tuler:
Can you add a `.env.sepolia` and a `.env.mainnet` with our aquarium configuration, just to make it easier for the developer?

tuler (approved):

PR #9, opened 2023-09-13
Chain selection

tuler:
Only 1 chain is supported, depending on the `NEXT_PUBLIC_CHAIN_ID` env var, as we will have one deployed per network.

I also added the connect button to the header, but we need to investigate how to integrate Mantine theme into the RainbowKit theme.
https://www.rainbowkit.com/docs/theming
https://www.rainbowkit.com/docs/custom-theme

ps: also bumped mantine to new beta.


brunomenezes:
Just need to add this env var on Vercel projects and I think it is all good.

tuler:
The button is very ugly.
Maybe I should hide it for now, and then research how to make it look good (follow Matine theme).

brunomenezes:
> The button is very ugly. Maybe I should hide it for now, and then research how to make it look good (follow Matine theme).

Yeah, since we are not using it. I agree that we are better off without it. Also, agree that it currently feels out of place.

tuler:
> Yeah, since we are not using it. I agree that we are better off without it. Also, agree that it currently feels out of place.

I made it look better, I think we can merge with it.

brunomenezes (approved):

PR #22, opened 2023-09-18
Package for generated wagmi hooks

tuler:
Create a package for the code generated by @wagmi/cli


packages/rollups-wagmi/wagmi.config.ts
```diff
@@ -0,0 +1,14 @@
...
+import hardhatDeploy from "@sunodo/wagmi-plugin-hardhat-deploy";
+
+export default defineConfig({
+    out: "src/index.tsx",
+    plugins: [
+        hardhatDeploy({
+            directory: "node_modules/@cartesi/rollups/export/abi",
+        }),
+        erc(),
+        react(),
```
tuler, line 12: It works without this line. There is some kind of typescript configuration problem for when react code is generated

PR #25, opened 2023-09-21
Feature: Input Details component

brunomenezes:
### Summary
Code changes to display the content from DApp outputs (Notices/Reports/Vouchers) that we are calling `InputDetails`. 

Other changes:
* Added Vitest and the testing-library to the UI package.
* Made all the requested changes from Prettier + Eslint to have a clean slate (that is way 43 changes).
* Added story + test cases for the UI components, including the `InputDetails`.
* Add a GitHub action to run lint and formatting checks, including the initial couple of tests added.


brunomenezes:
@tuler, 
I wondering if we could merge that PR. When I added the GitHub Action, I went through the rabbit hole of fixing all the prettier + Eslint requests to have a clean slate, but currently, that means 43 files were changed. 🫠

apps/web/codegen.ts
```diff
@@ -4,7 +4,7 @@ const schema = process.env.NEXT_PUBLIC_EXPLORER_API_URL;
 
 if (!schema)
     throw new Error(
-        "NEXT_PUBLIC_EXPLORER_API_URL environment variable is required for GraphQL code generation.",
+        "NEXT_PUBLIC_EXPLORER_API_URL environment variable is required for GraphQL code generation."
```
tuler, line 7: my editor is not formatting like this

tuler: A lot of other changes are related to this behavior.

.vscode/settings.json
```diff
@@ -0,0 +1,9 @@
+{
+    "editor.formatOnPaste": false,
```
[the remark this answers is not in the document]
brunomenezes: I add that because it usually adds extra line breaks.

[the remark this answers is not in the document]
tuler: I have to same effect as formatOnSave, just changes the trigger

apps/web/codegen.ts
brunomenezes: I fixed the prettier version, `2.8.8` instead of using `latest`, you probably have a different version installed because our `prettierrc` is very lean in terms of overrides. As I understood `prettier` rules can change from one version to another. To be honest, that causes more noise than helps. Hence, why I fixed the version. If we want to bump the version that will bring file changes alongside.

.prettierrc.json
```diff
@@ -1,5 +1,5 @@
 {
-    "tabWidth": 4,
+    "tabWidth": 4,
```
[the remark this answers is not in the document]
brunomenezes: I see there is a version 4. I will force-push a fix-up.

apps/web/codegen.ts
tuler: We should be using version 3

brunomenezes: That is fine, but I will fix the version. Also, whoever bumps the version also carries out fixing any of the "errors".

packages/ui/src/InputDetails/InputDetailsContext.tsx
```diff
@@ -0,0 +1,128 @@
...
+    availableContent: OptionalContents;
+};
+type Action = { type: "SET_DEFINED_CONTENT"; payload: OptionalContents };
+type Reducer = (state: State, action: Action) => State;
+
+type TAB = { label: string; icon: IconType; type: SupportedContent };
+
+const SUPPORTED_TABS = [
+    {
+        label: "Inputs",
```
tuler, line 31:
```suggestion
        label: "Input",
```

packages/ui/src/InputDetails/InputDetailsContext.tsx
```diff
@@ -0,0 +1,128 @@
...
+type State = {
+    supportedTabs: typeof SUPPORTED_TABS;
+    availableContent: OptionalContents;
+};
+type Action = { type: "SET_DEFINED_CONTENT"; payload: OptionalContents };
+type Reducer = (state: State, action: Action) => State;
+
+type TAB = { label: string; icon: IconType; type: SupportedContent };
+
+const SUPPORTED_TABS = [
```
tuler, line 29: I'd change the order to:
```
Input
Notices
Reports
Vouchers
```

packages/ui/src/InputDetails/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+                        <Text ref={ref} size={isSmallDevice ? "xs" : "md"}>
+                            {item.label}
+                        </Text>
+                    ),
+                );
+                const TabText = hasContent
+                    ? CText
+                    : () => (
+                          <Tooltip
+                              label={`${item.label} was not configured for that Input.`}
```
tuler, line 64:
```suggestion
                              label={`No ${item.label} were created on this input.`}
```

tuler (approved):

PR #30, opened 2023-09-28
Docker build and publishing

tuler:
Build a docker image with the nextjs application.
Followed the guidelines at https://turbo.build/repo/docs/handbook/deploying-with-docker and also https://nextjs.org/docs/pages/building-your-application/deploying#docker-image


apps/web/codegen.ts
```diff
@@ -1,11 +1,8 @@
 import type { CodegenConfig } from "@graphql-codegen/cli";
 
-const schema = process.env.NEXT_PUBLIC_EXPLORER_API_URL;
-
-if (!schema)
-    throw new Error(
-        "NEXT_PUBLIC_EXPLORER_API_URL environment variable is required for GraphQL code generation.",
-    );
+const schema = "https://squid.subsquid.io/cartesi-rollups-mainnet/graphql";
```
brunomenezes, line 3: @tuler, Is it not better to get from the environment variable instead of hardcoded?

docker-bake.hcl
```diff
@@ -0,0 +1,11 @@
+target "docker-metadata-action" {}
+target "docker-platforms" {}
+
+target "default" {
+  inherits   = ["docker-metadata-action", "docker-platforms"]
+  dockerfile = "docker/Dockerfile"
+  context    = "."
+  args = {
+    "ENVIRONMENT" = "sunodo"
```
brunomenezes, line 9: Are we going to generate every image with the environment `sunodo`?  I saw in the dockerfile that the environment there is `localhost`. So the idea is that for people to use it locally they need to build their own image, is that right?

apps/web/codegen.ts
tuler: the problem of using the environment is that often we want to build for localhost but the localhost environment is not necessarily running during the build

docker-bake.hcl
tuler: Yeah, this is kind of a hack, but it was the best I could do. I need the sunodo settings to run the image inside the sunodo environment.

apps/web/codegen.ts
brunomenezes: But then you can set up that build to whatever URL you want to work with. Because the environments (sepolia/mainnet) may be in parity now, but that is not guaranteed, and the testnet is more likely to have new graphQL updates that would cause build problems on Vercel, for example. I think we don't have this defined here on GitHub action variables, but I would setup to use Sepolia endpoints.

tuler: So maybe we should separate the `NEXT_PUBLIC_EXPLORER_API_URL` variable from the variable used by the codegen?

brunomenezes (approved):

apps/web/codegen.ts
brunomenezes: Yeah. For now,, it shouldn't be a problem and should be a blocker to proceed with Sunodo's developments. But we will need to improve that part for this code generator to be able to support multiple situations e.g. Our own Vercel deployments, docker image generation for external use, etc.

PR #35, opened 2023-10-08
feat: Add application and ERC-20 autocomplete fields to deposit form

nevendyulgerov:


brunomenezes (changes requested):
Hi @nevendyulgerov, 
After having a run on both Mainnet and Sepolia previews, for informational purposes for those who are using them, let's change a few things: 

1. Increase the severity of the alerts from Info(blue) to warning(yellow) for both the `Application` and  `ERC-20` inputs. The copies are good. _That is when the application does not exist, and the token is gonna be used for the first time._ 
2. Include a warning icon. We are using the [Tabler Icons](https://react-icons.github.io/react-icons/icons?name=tb) set. A good candidate would be the `TbAlertCircle`.

brunomenezes (approved):

Issue #20, opened 2023-09-15
Ether deposit


tuler:
## 📄 Context

DApps accept deposits of ether (chain native token)  by interacting with the [EtherPortal](https://next.docs.cartesi.io/cartesi-rollups/api/json-rpc/portals/EtherPortal/) smart contract.

## ✔️ Solution

Documentation about the Ether deposit:
https://next.docs.cartesi.io/cartesi-rollups/api/json-rpc/portals/EtherPortal/

The input for the application contract address could use an auto-complete by fetching the Application list from the backend, but this can be left for a future improvement (similar to https://github.com/cartesi/rollups-explorer/issues/19).

## 📈 Subtasks

- [ ] create the form for the deposit
- [ ] implement the depositEther transaction

## 🎯 Definition of Done

- [ ] transaction successful

PR #40, opened 2023-10-26
#20 Create form for depositing ether to selected dapp

nevendyulgerov:


packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,145 @@
...
+        </>
+    );
+};
+
+export interface EtherDepositFormProps {
+    applications: string[];
+    onSubmit: () => void;
+}
+
+export const EtherDepositForm: FC<EtherDepositFormProps> = (props) => {
```
tuler, line 67: Try to use `useForm`. Same comment as the one I left in PR #36

nevendyulgerov: Hi, @tuler , I utilised `useForm` as you requested.

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+    const loading =
+        deposit.status === "loading" || depositWait.status === "loading";
+
+    useEffect(() => {
+        if (depositWait.status === "success") {
+            form.reset();
+            onSubmit();
+        }
+    }, [depositWait.status, onSubmit]);
+
```
nevendyulgerov, line 69: We can ignore this warning, as adding `form` as dependency results in infinite resetting of the form when the deposit has succeeded.

tuler (commented):
I don't see the field for the ether amount to deposit

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+            application: (value) =>
+                value !== "" && isAddress(value) ? null : "Invalid application",
+            execLayerData: (value) =>
+                isHex(value) ? null : "Invalid hex string",
+        },
+        transformValues: (values) => ({
+            address: isAddress(values.application)
+                ? getAddress(values.application)
+                : "0x1",
+            hexExecLayerData: toHex(values.execLayerData),
```
tuler, line 45: call the field `execLayerData` instead of `hexExecLayerData`

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+        validate: {
+            application: (value) =>
+                value !== "" && isAddress(value) ? null : "Invalid application",
+            execLayerData: (value) =>
+                isHex(value) ? null : "Invalid hex string",
+        },
+        transformValues: (values) => ({
+            address: isAddress(values.application)
+                ? getAddress(values.application)
+                : "0x1",
```
tuler, line 44: use `zeroAddress` as default value instead of `0x1`

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+            hexExecLayerData: toHex(values.execLayerData),
+        }),
+    });
+    const { address, hexExecLayerData } = form.getTransformedValues();
+    const application = form.getInputProps("application");
+    const depositPrepare = usePrepareEtherPortalDepositEther({
```
tuler, line 50: name this var simply `prepare`

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+        args: [address, hexExecLayerData],
+        enabled: isAddress(application.value),
+    });
+    const deposit = useEtherPortalDepositEther(depositPrepare.config);
```
tuler, line 54: name this var `execute`

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+    const depositWait = useWaitForTransaction(deposit.data);
```
tuler, line 55: name this var simply `wait`

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+    const canSubmit =
+        form.isValid() &&
+        !depositPrepare.isLoading &&
+        depositPrepare.error === null;
+    const loading =
+        deposit.status === "loading" || depositWait.status === "loading";
+
+    useEffect(() => {
+        if (depositWait.status === "success") {
+            form.reset();
```
tuler, line 65: I don't think we need this logic of resetting

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+                    value={application.value}
+                    withAsterisk
+                    error={
+                        form.errors?.application ||
+                        (depositPrepare.error as BaseError)?.shortMessage
+                    }
+                    rightSection={
+                        depositPrepare.isLoading && <Loader size="xs" />
+                    }
+                    onChange={(application) => {
```
tuler, line 87: don't handle this manually, let the form handle it.
pass `{...form.getInputProps("application")}` to the `Autocomplete`

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+        <form>
+            <Stack>
+                <Autocomplete
+                    label="Application"
+                    description="The application smart contract address"
+                    placeholder="0x"
+                    data={applications}
+                    value={application.value}
+                    withAsterisk
+                    error={
```
tuler, line 80: we don't need this

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+    return (
+        <form>
+            <Stack>
+                <Autocomplete
+                    label="Application"
+                    description="The application smart contract address"
+                    placeholder="0x"
+                    data={applications}
+                    value={application.value}
```
tuler, line 78: don't handle this manually, let the form handle it.
pass {...form.getInputProps("application")} to the Autocomplete

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+import { useWaitForTransaction } from "wagmi";
+import { TransactionProgress } from "./TransactionProgress";
+
+export interface EtherDepositFormProps {
+    applications: string[];
+    onSubmit: () => void;
+}
+
+export const EtherDepositForm: FC<EtherDepositFormProps> = (props) => {
+    const { applications, onSubmit } = props;
```
tuler, line 28: I don't think we need this `onSubmit` callback

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+                    }
+                    onChange={(application) => {
+                        form.setFieldValue("application", application);
+                        form.clearFieldError("application");
+                    }}
+                    onBlur={() => application.onBlur()}
+                />
+
+                {!form.errors.application &&
+                    isHex(application.value) &&
```
tuler, line 95: Use `address != zeroAddress`

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+                    !applications.includes(application.value) && (
```
tuler, line 96: use `address`

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+                        <Alert
+                            variant="light"
+                            color="yellow"
+                            icon={<TbAlertCircle />}
+                        >
+                            This is an undeployed application.
+                        </Alert>
+                    )}
+
+                <Textarea
```
tuler, line 106: We should hide this under an `Advanced` option, like I did in the ERC-20 deposit form

nevendyulgerov:
> I don't see the field for the ether amount to deposit

@tuler I'm not sure how to pass the amount to the deposit hook. Looking at https://next.docs.cartesi.io/cartesi-rollups/api/json-rpc/portals/EtherPortal/#depositether I don't see information about the deposit amount. Can you please elaborate how this can be done?

CC @brunomenezes

packages/ui/src/EtherDepositForm.tsx
nevendyulgerov: @tuler , I'm using the `onSubmit` callback from the `shell` to refresh the page on successful deposit so that the homepage table get's updated accordingly. If you think this is not required, I'll remove this logic.

packages/ui/src/EtherDepositForm.tsx
nevendyulgerov: @tuler , this won't work as expected because the applications array contains addresses in format `0x60a7048c3136293071605a4eaffef49923e981cc` while the `getAddress` util returns addresses in format `0x60A7048C3136293071605A4eAFfeF49923E981cc`. So the check `applications.includes(address)` will always return `false` and present the alert even when you select an application from the list. Here's a video demonstrating this:

https://github.com/cartesi/rollups-explorer/assets/6005179/3176c26d-b5b2-4c57-9a0c-9cba79fd0a6b

packages/ui/src/EtherDepositForm.tsx
nevendyulgerov: @tuler , with this logic on successful deposit, the form's inputs get cleared, which is a standard on submit behaviour. Still, if you think we don't need it, I'll remove this logic.

packages/ui/src/EtherDepositForm.tsx
nevendyulgerov: @tuler can you please elaborate why we don't need an error for the applications autocomplete? Looking at `ERC20DepositForm` I see we have similar error handling for `TokenAutocomplete`. Also, since sepolia requests are failing a lot with 503 errors, keeping the `error` field allows users to see those errors.

packages/ui/src/EtherDepositForm.tsx
tuler: Yeah, I don't think it's required

packages/ui/src/EtherDepositForm.tsx
tuler: You have a two options. Compare ignoring case. Or convert everything to checksummed address using `getAddress`

apps/web/src/app/shell.tsx
```diff
@@ -12,18 +15,22 @@ import {
...
 import { ConnectButton } from "@rainbow-me/rainbowkit";
 import Link from "next/link";
-import { FC, ReactNode } from "react";
 import { TbApps, TbHome, TbMoonStars, TbPigMoney, TbSun } from "react-icons/tb";
 import { useAccount } from "wagmi";
-import CartesiLogo from "../components/cartesiLogo";
 import { useApplicationsQuery, useTokensQuery } from "../graphql";
 
 const Shell: FC<{ children: ReactNode }> = ({ children }) => {
+    const router = useRouter();
```
tuler, line 28: this variable is unused

apps/web/src/app/shell.tsx
```diff
@@ -12,18 +15,22 @@ import {
     useMantineTheme,
 } from "@mantine/core";
 import { useDisclosure } from "@mantine/hooks";
+import { FC, ReactNode } from "react";
+import CartesiLogo from "../components/cartesiLogo";
```
tuler, line 19: please organize imports

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,183 @@
...
+            amount: (value) =>
+                value !== "" && Number(value) > 0 ? null : "Invalid amount",
+            execLayerData: (value) =>
+                isHex(value) ? null : "Invalid hex string",
+        },
+        transformValues: (values) => ({
+            address: isAddress(values.application)
+                ? getAddress(values.application)
+                : zeroAddress,
+            amountBigint:
```
tuler, line 64: you can call the variable `amount` only

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,183 @@
...
+                values.amount !== "" ? BigInt(values.amount) : undefined,
```
tuler, line 65: this conversion from the filled string to BigInt is not correct. You have to convert using viem `parseUnits`.
The number of decimals is given ideally by querying the native currency of the connected chain.

```typescript
const { chain } = useNetwork();
...
parseUnits(values.amount, chain?.nativeCurrency.decimals ?? 18)
```

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -0,0 +1,183 @@
...
+    const wait = useWaitForTransaction(execute.data);
+    const canSubmit =
+        form.isValid() && !prepare.isLoading && prepare.error === null;
+    const loading = execute.status === "loading" || wait.status === "loading";
+
+    useEffect(() => {
+        if (wait.status === "success") {
+            form.reset();
+        }
+    }, [wait.status]);
```
tuler, line 86: So add the form here to get rid of the warning

nevendyulgerov: The form doesn't have a stable reference and this will result in infinite re-invocation of that effect.

tuler (approved):

packages/ui/test/EtherDepositForm.test.tsx
```diff
@@ -0,0 +1,197 @@
+import { render, screen, fireEvent } from "@testing-library/react";
+import { afterAll, describe, it } from "vitest";
+import { EtherDepositForm } from "../src/EtherDepositForm";
+import withMantineTheme from "./utils/WithMantineTheme";
+import { useNetwork } from "wagmi";
```
dandheedge, line 5: This import is unused, I think we can remove it

dandheedge (changes requested):

packages/ui/test/EtherDepositForm.test.tsx
nevendyulgerov: Good catch, I addressed this, @dandheedge .

dandheedge (approved):
LGTM!

PR #41, opened 2023-10-30
Feature/dapp connect

brunomenezes:
### Summary
Code changes to add a feature to create and persist a connection to get data from a DApp's backend (graphQL API). Load details about the input types, i.e. [`notices`, `vouchers`, `reports`]. 

The idea was to use what the library provides rather than add a dependency (e.g. Redux, Jotai, Zustand). So, the layer where react interacts more intensively is a combination of `Context + Reducer + Hooks`; for the lower layer, I defined an interface called `repository` that establishes a contract to be implemented, adding the `persistent` feature (For this version the first repository implementation uses the LocalStorage, unsurprisingly named `localRepository`) that open possibility to other repository implementations like one for Sunodo as the provider is ready to accept a custom repository if needed.

I also use the concept of `View`, meaning all the connection-related components have everything at hand by using hooks as far as it is under the provider (at the very top). The provider also holds the creation modal and ways to interact with it through hooks, so there are no prop-drilling nor disconnected modal+forms around. Everything is colocated and intentional.

I also updated the graphQL **Applications** query to include the limit variable. Since we have cardinality defined in our backend graphQL schema setting reasonable limits (you can read more about [here](https://docs.subsquid.io/graphql-api/dos-protection/)) and if a query reaches that limit of "points", it will throw errors, so list everything is not a good way to go. With that, I list the first ten applications and, debounce the inputs, check for likely values and bring the first ten results that matched, which also gave the possibility to warn the user when adding an address that we did not index (meaning it is a theoretical undeployed application).

That is the gist. I picked and fixed other things along the way that were a bit out of the scope but necessary nonetheless (e.g. On small screens, the `Connect Wallet` was not visible anywhere)

**PS: I had problems writing tests for one component only, `<InputDetailsView />`, and I created the following issue #48. It went down a rabbit hole, so it's better to work in a clean slate focused solely on that.**


tuler (commented):
The style of the `Connect` button in the notice panel is weird. The font size looks off from the standard.

brunomenezes:
> The style of the `Connect` button in the notice panel is weird. The font size looks off from the standard.

I removed the `compact-lg` on [e85451b](https://github.com/cartesi/rollups-explorer/pull/41/commits/e85451bcf20f99d35fdc213b25d816ea3995b11c)

tuler:
Screenshot 2023-11-09 at 12 43 38 AM

The disposition of these buttons is strange. Two options:

1) put the "Test Connection" button on the left side of the "Save" button.
2) remove the "Test Connection" button and do the test action on change of the url value (with some debounce).

As we have this test feature, which is nice, the "Save" button should only be enabled if test is successful.

Screenshot 2023-11-09 at 12 47 29 AM

The result of the test can be improved as well. That display is strange.

tuler:
I feel the drawer is diverging from the application general behavior.
How about we put the connection info as a column of the applications table at http://localhost:3000/applications

brunomenezes:
> Screenshot 2023-11-09 at 12 43 38 AM
> The disposition of these buttons is strange. Two options:
> 
> 1. put the "Test Connection" button on the left side of the "Save" button.
> 2. remove the "Test Connection" button and do the test action on the change of the URL value (with some debounce).
> 
> As we have this test feature, which is nice, the "Save" button should only be enabled if the test is successful.
> 
> Screenshot 2023-11-09 at 12 47 29 AM
> The result of the test can be improved as well. That display is strange.

My idea is to be permissive on that `connection test` piece, and the one adding the connection can decide to test it. It is more of a feedback for the user than for the form. 

I will make some adjustments.

brunomenezes:
> I feel the drawer is diverging from the application general behavior. How about we put the connection info as a column of the applications table at http://localhost:3000/applications

Adding information in the application's table is fine; it is an excellent "shortcut" if you include actions there, too. But the drawer is essential to hold configurations in an organized place.

tuler (approved):
Looks good!

PR #42, opened 2023-11-01
#39 Improve ERC20Portal data display

### LINKED ISSUE #39: Improve ERC20Portal data display
## 📄 Context
We have knowledge and control over how to parse the encoded content of an **ERC20Portal deposit**. Therefore, an improvement can be made to display the content of this encoded payload in the sections `As Text` and `As JSON`. 

## ✔️ Solution
A code reference to the solidity code that does the encoding: [InputEncoding.sol#encodeERC20Deposit](https://github.com/cartesi/rollups-contracts/blob/73cee6a4b8b7d6874601dd2c6fa930e0dba39964/onchain/rollups/contracts/common/InputEncoding.sol#L40)
ERC20Portal ABI reference for the function: [depositERC20Tokens](https://github.com/cartesi/rollups-contracts/blob/main/onchain/rollups/deployments/sepolia/ERC20Portal.json#L15)
Leverage the Ethereum utility library [viem](https://viem.sh/docs/getting-started.html) that is available in the project.

The decoded version could become an object with the same name fields and easily be consumed by the sections mentioned as JSON. 

_PS: Because it is `encodePacked` the decoding needs to be done manually; it is not just drop the ABI. But it is still worth noting the field names in the refs above._

## 📈 Subtasks

- [ ] Decode the ERC20Portal input payload
- [ ] Display the decoded content `As Text`
- [ ] Display the decoded content `As JSON`

## 🎯 Definition of Done

- [ ] The ERC20Portal payload content is now human-readable.

## 📸 Snapshots (extra)

Input row level content
Screenshot 2023-10-21 at 12 25 12

Data sections
Screenshot 2023-10-21 at 12 25 22

dandheedge:
Close #39


nevendyulgerov (approved):

apps/web/src/components/inputRow.tsx
```diff
@@ -46,6 +55,19 @@ const methodResolver: MethodResolver = (input) => {
     }
     return undefined;
 };
+export const decodePayload = (payload: Hex): string => {
+    const isERC20Method = slice(payload, 0, 1) === "0x01";
```
tuler, line 59: This logic is not correct. It's not the first byte as `0x01` that indicates the payload is a ERC-20 deposit. It's the sender of the input.

brunomenezes (changes requested):
There are some requested changes to be addressed as pointed by @tuler

dandheedge:
> There are some requested changes to be addressed as pointed by @tuler

@brunomenezes I added the second params on `decodePayload` to check whether the sender address of the input is the  `erc20PortalAddress`

tuler (commented):
Instead of doing a specific change for ERC-20 deposits I'd rather handle this in the scope of the more general problem of how to present different kinds of binary data.

PR #51, opened 2023-11-07
Refactor: Packages bundling setup

brunomenezes:
### Summary
Code changes to the bundling setup for both rollups-wagmi and ui packages. Also include some refactoring to push client components down the component's tree. 

Changes:
* Add the `use client;` directive for UI and rollups-wagmi bundling. Also, I described in the `package.json` what source applications will use when using `import` or `require`, which solves the problem for Vitest. 
* I made the `layout.tsx` a server component again and grouped the providers in a client component, pushing client components a bit down the tree. You can read a bit more about why that is recommended in the composition pattern page in the session [here](https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#moving-client-components-down-the-tree)
* I included test cases for the missing component `<InputDetailsView />`. Also, I did some refactoring in the `test/utils/**` and added small bits of code in the `<InputDetails />` component in the UI package to make testing a breeze.
* **In dev mode, there are warnings regarding differences between server and client regarding the RainbowKit theme setup, but only in case you start the application on light mode instead of dark. Nonetheless, it is nothing to worry about. Another thing that may fire a warning on that hydration space is the use of browser plugins. In my case, Grammarly injects properties in the `<body>` tag, causing react/nextJS to warn about mismatches in the component tree.**


packages/ui/src/InputDetails/PageableContent.tsx
```diff
@@ -129,6 +130,9 @@ export const PageableContent: FunctionComponent<PageableContentProps> = ({
                                             disabled={pageNumber === 1}
                                             aria-label="Button to load the previous content"
                                         >
+                                            <VisuallyHidden>
```
tuler, line 133: Why do you need this?

brunomenezes: That affects two things. It is available to screen readers and facilitate testing since it is something users would interact with. That would follow the motto of react-testing-library. The other way would be plug `data-testid` that is a escape-hatch to be used with the testing library I mentioned.

tuler (approved):
I didn't understand why package/ui you did .mjs and rollups-config you did .cjs

brunomenezes:
> I didn't understand why package/ui you did .mjs and rollups-config you did .cjs

You mean wagmi config? Is just because wagmi is type module so tsup generate .js as ESM and CJS becomes the “second class”. The ui I did not touch that and rather only did the exports mapping.

PR #54, opened 2023-11-10
Improve allowance/balance UI on amount field for deposit ERC-20 Tokens

### LINKED ISSUE #43: "Max amount" field is confusing
## 📄 Context

On the deposit form, there is a text field labeled "Max amount".
This may confuse the user into thinking they can edit it for some reason.

Screenshot from 2023-11-02 07-50-07

## ✔️ Solution

Make the maximum amount be displayed through a read-only component, like a text component.
Preferably, this would sit near to the "Amount" field, that comes right after it.

dandheedge:
Close #43

### DIFF
--- a/packages/ui/src/ERC20DepositForm.tsx
+++ b/packages/ui/src/ERC20DepositForm.tsx
@@ -126,6 +126,7 @@ export const TokenAutocomplete: FC<TokensAutocompleteProps> = (props) => {
                     );
                     onChange(formattedValue);
                 }}
+                data-testid="token-input"
             />
 
             {erc20Address !== "" &&
@@ -184,7 +185,6 @@ export const ERC20DepositForm: FC<ERC20DepositFormProps> = (props) => {
                   return (d.error as BaseError).shortMessage;
               })
         : [];
-
     // token amount to deposit
     const [amount, setAmount] = useState<string>("");
     const amountBigint =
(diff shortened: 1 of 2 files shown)


packages/ui/src/ERC20DepositForm.tsx
```diff
@@ -301,9 +286,46 @@ export const ERC20DepositForm: FC<ERC20DepositFormProps> = (props) => {
                             rightSectionWidth={60}
                             rightSection={<Text>{symbol}</Text>}
                             onChange={(e) => setAmount(e.target.value)}
+                            value={amount}
                             withAsterisk
                             data-testid="amount-input"
                         />
+                        <Flex
+                            mt={"-sm"}
```
nevendyulgerov, line 294: Static props like `mt` in this case can be expressed using the following syntax:

```
mt="-sm"
```

No need for curly brackets here. Please reformat the code below using this approach.

packages/ui/src/ERC20DepositForm.tsx
```diff
@@ -301,9 +286,46 @@ export const ERC20DepositForm: FC<ERC20DepositFormProps> = (props) => {
...
+                            justify={"space-between"}
+                            direction={"row"}
+                        >
+                            <Flex rowGap={6} c={"dark.2"}>
+                                <Text fz="xs">Balance:</Text>
+                                <Text fz="xs" mx={4}>
+                                    {" "}
+                                    {balance != undefined && decimals
```
nevendyulgerov, line 302: Please use strict comparison operator. `!==` instead of `!=`. With it, you check the value and its type as well. I know that in this component there are other instances where "loose" comparison operators are used. Those were refactored in https://github.com/cartesi/rollups-explorer/pull/52/files.

packages/ui/src/ERC20DepositForm.tsx
```diff
@@ -301,9 +286,46 @@ export const ERC20DepositForm: FC<ERC20DepositFormProps> = (props) => {
...
+                                        ? formatUnits(balance, decimals)
```
nevendyulgerov, line 303: I think we should append here the token symbol.

packages/ui/src/ERC20DepositForm.tsx
```diff
@@ -301,9 +286,46 @@ export const ERC20DepositForm: FC<ERC20DepositFormProps> = (props) => {
...
+                                        Max
+                                    </UnstyledButton>
+                                )}
+                            </Flex>
+                            <Flex rowGap={6} c={"dark.2"}>
+                                <Text size="xs">Allowance:</Text>
+                                <Text ml={4} fz="xs">
+                                    {allowance != undefined && decimals
+                                        ? formatUnits(allowance, decimals)
+                                        : ""}
```
nevendyulgerov, line 325: Same as above, we should append the token symbol here.

packages/ui/src/ERC20DepositForm.tsx
```diff
@@ -301,9 +286,46 @@ export const ERC20DepositForm: FC<ERC20DepositFormProps> = (props) => {
...
+                        >
+                            <Flex rowGap={6} c={"dark.2"}>
+                                <Text fz="xs">Balance:</Text>
+                                <Text fz="xs" mx={4}>
+                                    {" "}
+                                    {balance != undefined && decimals
+                                        ? formatUnits(balance, decimals)
+                                        : ""}
+                                </Text>
+                                {balance != undefined && decimals && (
```
nevendyulgerov, line 306: Shouldn't the `Max` button appear only if balance is larger than zero? Currently, it's displayed even if balance is zero:

Screenshot 2023-11-17 at 12 31 05

nevendyulgerov (commented):
Hey @dandheedge , I left some comments for you to consider.

Please also add unit tests for the new UI.

dandheedge:
> Hey @dandheedge , I left some comments for your to consider.
> 
> Please also add unit tests for the new UI.

Sure @nevendyulgerov. I will update it.

packages/ui/src/ERC20DepositForm.tsx
dandheedge: @nevendyulgerov I think we don't need the token symbol at the bottom because it's already displayed on the right side of the input. For instance, Uniswap shows the token symbol just once, either on token selection or when the ERC20 Address is pasted.

Screenshot 2023-11-21 at 12 04 12

Screenshot 2023-11-21 at 16 00 33


cc @brunomenezes also to give some insight

nevendyulgerov (approved):
LGTM, thanks for making those tweaks, @dandheedge !

packages/ui/src/ERC20DepositForm.tsx
brunomenezes: I agree on keeping it as lean as possible. Since the token symbol is already inside the input and both balance and allowance belong to the Input in question, repeating the symbol is redundant.

brunomenezes (approved):

Issue #53, opened 2023-11-09
Improvements for forms


nevendyulgerov:
Improvements for forms, suggested by @brunomenezes :

I have two things to mention here. I agree with @tuler about the header shell fetching this kind of data. I think you could have a `<Deposit />` that encapsulates all the data needs of the form/forms (see where I am getting here? based on the suggestion on #44 ). The other thing that needs to be addressed is that `useApplicationsQuery`. Because requests with no limits will soon reach the imposed cardinalities in the backend, that will return errors like `response might exceed the size limit`. I also worked with application IDs and mentioned what I did on my current open PR #41. 

So my suggestions are : 

* Encapsulate your needs in a separate Component using the fetching data, modal and form/forms inside it
* look in my PR and adjust the query graphQL (that will come with my PR merged anyway) and use the limit for requesting only X amount of items to be more future-proof.

_Originally posted by @brunomenezes in https://github.com/cartesi/rollups-explorer/pull/36#discussion_r1387348732_

brunomenezes:
Closing since the #55 was merged into #36.

Issue #44, opened 2023-11-02
All-in-one deposit button


guidanoli:
## 📄 Context

Currently, AFAIK the production version of the explorer only supports ERC-20 deposits,
but it seems also that you are in the process of supporting Ether deposits as well.
Later, you might also give support to ERC-721 tokens (NFTs) and ERC-1155 tokens (Multi tokens).
My fear is that if all of these assets have their own individual deposit button, it might overcrowd the UI.

## ✔️ Solution

To solve this issue, we can make an all-in-one deposit button, which, when clicked, prompts the user to choose the specific type of asset they would like to deposit. It can be a pop-up component with buttons for each asset type, or a drop-down list streaming down from the deposit button...

nevendyulgerov:
I think @guidanoli has a good point here. Hey, @brunomenezes , @tuler if you are ok with the suggested solution, I can start working on it. Please let me know.

brunomenezes:
> I think @guidanoli has a good point here. Hey, @brunomenezes , @tuler if you are ok with the suggested solution, I can start working on it. Please let me know.

That is interesting, @guidanoli. Are these standards already in use on the DApp level? Is there a Dapp you tried to use one of these? 

I have been talking to @tuler about that crowded header, which we want to avoid having. For example, the create connection that is coming with Dapp-connect #41 is introducing a drawer where configs would live (e.g. there is a create custom ABI we'll experiment with soon that will also use that space). 

Nonetheless, that idea is not odd to me; when I worked on International Payments, the form would change based on (currency + country + specific bank country nuances) 

@nevendyulgerov, first, we need to get the following: #49 and #40 approved and merged before working on a new design.

guidanoli:
> Are these standards already in use on the DApp level? Is there a Dapp you tried to use one of these?

I think Ether and ERC-20 are, by far, the most widely used types of assets. There are, however, some DApp that deal with NFTs (ERC-721 tokens). [Auction DApp](https://github.com/cartesi/rollups-examples/tree/main/auction) and [Tribes](https://github.com/felipefg/tribes) are some examples.

As for ERC-1155 tokens (multi-tokens), I don't know any example that uses them yet, but they are interesting for transferring tokens of different types in bulk. For example, imagine of an RPG where you have several resources (grain, ore, lumber, wool, clay). With ERC-1155, you can transfer resources of different types in a single transaction.

juztamau5:
[...]

PR #55, opened 2023-11-10
#53 Add SendTransaction component for holding all transaction-related forms in a single place

nevendyulgerov:


brunomenezes:
Hey! I have a suggestion seeing that change. Because the `Raw Input` is not a deposit and when I suggested to have a "<Deposit />" component I was think more of the `ERC20 deposit` and `ETH deposit` under that component given the discussion [here](https://github.com/cartesi/rollups-explorer/issues/44#issuecomment-1803144650). But playing around with it I think we could change the **copy** and also component holder name. 

We could change the `Deposit` button to `Send Transaction` or just `Send` using a different icon like the [TbArrowsDownUp](https://react-icons.github.io/react-icons/search?q=TBArrowsDownUp). 
send-transaction-header-button

The title changing from `Deposit` to `Send Transaction`, and inside the select we could use the [Group Options](https://mantine.dev/core/select/#group-options) having a `Deposit` group that includes the ERC20 and later coming the `ETH deposit` (that is in a separate branch) and finally `Others` with the `Raw Input` as an option in that group.
modal-header-select-content

Because they are different things but the high-level abstraction is that they are `transactions` of some sort (deposit / input / etc).

What do you think @tuler , @nevendyulgerov ?

tuler:
I think it could be a "Send Transaction" button, which opens a dialog.
The dialog could use a [SegmentedControl](https://mantine.dev/core/segmented-control/) similar to what I did below.
bridge

The options would be:
- Ether Deposit
- ERC-20 Deposit
- ERC-721 Deposit
- ERC=1155 Deposit
- Address Relay
- Raw Input

Maybe 6 options will be too much for a SegmentedControl. Need to test.

brunomenezes:
> I think it could be a "Send Transaction" button, which opens a dialog. The dialog could use a [SegmentedControl](https://mantine.dev/core/segmented-control/) similar to what I did below. bridge
> 
> The options would be:
> 
> * Ether Deposit
> * ERC-20 Deposit
> * ERC-721 Deposit
> * ERC=1155 Deposit
> * Address Relay
> * Raw Input
> 
> Maybe 6 options will be too much for a SegmentedControl. Need to test.

@tuler,  I knew you would mention the segmented control, but for 2 or 3 options with short text, I guess it would look nice but is not just about the number of options but the string length inside it too. I think the good/old Select is the way to go using that presentation of `Groups`. Because it is not that you will be quickly switching through them like in the `<InputDetails />`.

nevendyulgerov:
Hi @brunomenezes and @tuler , yes I agree that the copy was not great. That also caught my attention but initially I proceeded with it so I can get something out and then collect feedback from you guys. I'll work on the copy and suggested UI. Thanks for looking into it!

nevendyulgerov:
Hi @brunomenezes, @tuler , I made some changes to this branch, so that both suggestions (for Select with grouping and SegmentedControl) were taken into consideration.

Here's how the SendTransaction component looks with `SegmentedControl`:

https://github.com/cartesi/rollups-explorer/assets/6005179/fa668f50-4f17-4bd6-824b-b79785c4cefd

And here's how it looks with `Select` with groupings:

https://github.com/cartesi/rollups-explorer/assets/6005179/d26ae03d-d918-4e34-821b-2e0a29b0271f

I think both UIs look fine. A couple of thoughts:

- `SegmentedControl` UI looks ok as long as we are not going to have more options in that component. It's a bit easier to switch between the different forms with it than with a `Select`.
- `Select` UI is the better option if we are going to have more options down the road.

brunomenezes:
> Hi @brunomenezes, @tuler , I made some changes to this branch, so that both suggestions (for Select with grouping and SegmentedControl) were taken into consideration.
> 
> Here's how the SendTransaction component looks with `SegmentedControl`:
> 
>  Screen.Recording.2023-11-16.at.14.57.09.mov 
> And here's how it looks with `Select` with groupings:
> 
>  Screen.Recording.2023-11-16.at.14.56.07.mov 
> I think both UIs look fine. A couple of thoughts:
> 
> * `SegmentedControl` UI looks ok as long as we are not going to have more options in that component. It's a bit easier to switch between the different forms with it than with a `Select`.
> * `Select` UI is the better option if we are going to have more options down the road.
> 
> I added stories under `apps/workshop` storybook. They are located in `SendTransaction` and demonstrate the two UIs.

The `SegmentControl` is based on Apple's [equivalent](https://developer.apple.com/design/human-interface-guidelines/segmented-controls#Best-practices). IMHO it is suitable for fewer options like between 2 or 3 items considering the label content to be used to keep it `coherent`. The `Select` is a better option because we will probably work with more forms and not forget about different screen sizes (Click the gif below if it is too small). You can set it to be "vertical" for sure, but it still becomes noisy, and it is not the focus of this UI as a whole, but it ended up winning for the sheer amount of content displayed. That is my 2c here. What do you think @tuler ?

nevendyulgerov:
Hey @tuler , please give this one another review when you can. We explored two possible UIs for the form activation - `Select` or `SegmentedControl`.

tuler:
The select definitely looks better in this case.

I'd change the values to be more precise, instead of `ERC-20` use `ERC-20 Deposit`

nevendyulgerov:
I think this one is good to go, please review it when you can @brunomenezes , @tuler , @dandheedge .

dandheedge:
> I think this one is good to go, please review it when you can @brunomenezes , @tuler , @dandheedge .

Hi @nevendyulgerov ,

I noticed a minor issue regarding the margin style for the alert; it seems slightly off. Apart from that, everything looks good.

Screenshot 2023-12-02 at 16 47 32

dandheedge:
> > Hi @nevendyulgerov ,
> > I noticed a minor issue regarding the margin style for the alert; it seems slightly off. Apart from that, everything looks good.
> > Screenshot 2023-12-02 at 16 47 32
> 
> Hi, @dandheedge , I'm unable to reproduce this one. This is what I'm seeing in this case: Screenshot 2023-12-04 at 18 09 14

Hi @nevendyulgerov , yes that is exactly what I'm talking about. If you see the text _This is the deposit to an undeployed application_ and _This is the first deposit of the token_ you could see that the text is not align with the icon on the left

nevendyulgerov:
Hey @dandheedge , sorry for the confusion. You were right about the alert. This turned out to be a regression coming from mantine `v7.2.x` (we recently upgraded to it). There's an issue about this in their repo - https://github.com/mantinedev/mantine/issues/5305. I upgraded the library to `v7.3.x` and now the alert is displayed correctly:
Screenshot 2023-12-05 at 9 21 14

brunomenezes (approved):
Hi @nevendyulgerov, It looks good. 
I pushed a few updates. The **first** was for changing the send-transaction icon; it got lost in the conversation. The **second** was the type casting for the `descend()` function, replacing it with a `Type` param. I don't know why you need to add all this casting. 

From
```javascript
   descend(
        prop<number>("timestamp" as unknown as number) as unknown as (
            obj: Connection,
        ) => Ord,
    )
```

To:
```
  descend<Connection>(prop<number>("timestamp"))
```

brunomenezes (approved):
Hi @nevendyulgerov, It looks good. 
I pushed a few updates. The **first** was for changing the send-transaction icon; it got lost in the conversation. The **second** was the type casting for the `descend()` function, replacing it with a `Type` param. ~I don't know why you need to add all this casting.~ **UPDATE**: after yarn-install, the red squiggles show up and the build that was previously passing started failing on type checking. I can't point out what may change on yarn installs affecting that step on the NextJS build.

From
```javascript
   descend(
        prop<number>("timestamp" as unknown as number) as unknown as (
            obj: Connection,
        ) => Ord,
    )
```

To:
```
  descend<Connection>(propOr<number>(0, "timestamp"))
```

PR #64, opened 2023-11-19
Chore: Upgrade Mantine through workspaces

brunomenezes:
### Summary 
Upgrade Mantine from `7.0.0` to the latest `7.2.2`.

Changes: 
* Upgrade Mantine on the web, workshop and @cartesi/rollups-explorer-ui.
* Fix typescript checks on the UI package.
* Add two rules explicitly even though using the `default value`. Because, oddly enough, on the CI or locally, when running the `yarn format:check` throw warnings and applying the  `format` task changes a lot of files, not respecting the defaults 🫠


.prettierrc.json
```diff
@@ -1,5 +1,7 @@
 {
     "tabWidth": 4,
+    "trailingComma": "all",
+    "bracketSpacing": true,
```
tuler, line 4: You sure you need to add these two settings?

brunomenezes: It is what I explained above. Oddities that broke the CI and also when running the `yarn format:check` locally display the need to remove trailing commas as an example, even though the default value is to `add` the comma. So, I am just explicitly adding these.

nevendyulgerov (approved):
LGTM!

Issue #62, opened 2023-11-15
Rearrange navigational links


brunomenezes:
## 📄 Context
The header is becoming crowded with navigation and call-to-action kind of elements. Let's make the navigation part in a separate application area based on screen sizes, i.e. a Navbar.

## ✔️ Solution
By moving page links from the header to the existing Navbar. Also, updating the current Navbar behaviour; 

* Adjust the Navbar to maintain the current behaviour/content for small screens, but it becomes always visible for big screens.
* Remove the links [`Home`, `Applications`] from the header to the Navbar. 
* You may need to adjust the default width for the desktop in case the area feels too big for the content. _(check the theme object for that)_
* Header should keep the `Connect Wallet`, `Deposits`, `dark/light mode` and `dots` buttons on a big screen.
* The burger icon continues to only show up on small screens.
* The Navbar should be collapsed by default on small screens.

## 📈 Subtasks

- [x] Navbar is always visible on **big screens.**
- [x] Navbar is collapsed by default on **small screens.**
- [x] Navbar displays the current connect / deposit actions on **small screens.**
- [x] The links were moved from the header to the NavBar on **big screens.**

## 🎯 Definition of Done

- [x] The wallet-related actions and navigational links are visible on small screens.
- [x] The Navbar only displays the navigational links on big screens.

brunomenezes:
There is PR that may affect the changes here. So keep a eye on https://github.com/cartesi/rollups-explorer/pull/61

PR #66, opened 2023-11-20
#62 Display navbar on desktop and reorganise header links and buttons

nevendyulgerov:


apps/web/test/app/shell.test.tsx
```diff
@@ -120,4 +120,57 @@ describe("Shell component", () => {
             expect(logoLink.getAttribute("href")).toBe("/");
         });
     });
+
+    describe("Header", () => {
+        it("should display deposit links in header", () => {
+            const { container } = render(<Component>Children</Component>);
+            const header = container.querySelector("header") as HTMLDivElement;
```
brunomenezes, line 127: @nevendyulgerov, 
Since you are using the `data-testid` property, You could use the `screen` from the `@testing-library/react` and use the methods available, then you can even change the `expect`. The testing-library motto is to test what the user sees, using the low-level (query-selector) or data-testid property as an escape hatch for situations where something else is needed for any reason.


For reference: Look at the tests inside the `components/` folder. They're more idiomatic.

apps/web/test/app/shell.test.tsx
```diff
@@ -120,4 +120,57 @@ describe("Shell component", () => {
...
+            const depositButton = header.querySelector(
+                '[data-testid="deposit-button"]',
+            );
+            const depositEtherButton = header.querySelector(
+                '[data-testid="deposit-ether-button"]',
+            );
+
+            expect(depositButton !== null).toBe(true);
+            expect(depositEtherButton !== null).toBe(true);
```
brunomenezes, line 136: As mentioned above, using one of the `screen` methods, you could also change the expectation to something like `toBeInTheDocument()` instead of asserting against boolean expressions.

apps/web/test/app/shell.test.tsx
```diff
@@ -120,4 +120,57 @@ describe("Shell component", () => {
...
+        });
+
+        it("should not display home and applications links in header", () => {
```
brunomenezes, line 139: for this test case same as above.

apps/web/test/app/shell.test.tsx
```diff
@@ -120,4 +120,57 @@ describe("Shell component", () => {
...
+                '[data-testid="applications-link"]',
+            );
+
+            expect(homeLink === null).toBe(true);
+            expect(applicationsLink === null).toBe(true);
+        });
+    });
+
+    describe("Navbar", () => {
+        it("should display home and applications links in navbar", () => {
```
brunomenezes, line 153: Same as above, but if narrowing the document nodes is necessary, check inside the test folder `components/. You will find examples of narrowing the search and using the detached screen methods.

brunomenezes (changes requested):

brunomenezes (approved):

PR #69, opened 2023-11-21
56. Fix the logical condition for displaying the warning.

### LINKED ISSUE #56: Deposit wrongly displaying warnings
## 🙂 Expected behavior

When copying/pasting the Address of an existing application or already used ERC20 token, it should not display a warning message stating either `This is a deposit to an undeployed application` or `This is the first deposit of that token.`

## 🫠 Actual behavior

It does not match the address with ones already indexed by CartesiScan and keeps showing the warning message, forcing the user to click in the `autocomplete` options even though the Address is correct.

## 🧪 Minimal test case

Using the Sepolia network, click the  `Deposit` button and use the following addresses in the correct fields.

Application: 0x4cA2f6935200b9a782A78f408F640F17B29809d8 (HoneyPot)
ERC20 contract address: 0x059c7507B973d1512768c06f32a813BC93D83EB2 (SIM20)

## 🌎 Environment

**Sepolia/Mainnet** they are both affected.

## ✔️ Possible solutions

This is probably a matter of a `String case` comparison mismatch. So, leveraging the `Viem` [isAddressEqual()](https://viem.sh/docs/utilities/isAddressEqual.html#isaddressequal) function should put that problem to rest.

dandheedge:
Close #56

### DIFF
--- a/packages/ui/src/ERC20DepositForm.tsx
+++ b/packages/ui/src/ERC20DepositForm.tsx
@@ -287,7 +287,7 @@ export const ERC20DepositForm: FC<ERC20DepositFormProps> = (props) => {
                             nextValue.indexOf("0x"),
                         );
                         form.setFieldValue("erc20Address", formattedValue);
-                        onSearchTokens(nextValue);
+                        onSearchTokens(formattedValue);
                     }}
                 />
 
--- a/packages/ui/src/ERC20DepositForm.tsx
+++ b/packages/ui/src/ERC20DepositForm.tsx
@@ -49,7 +49,7 @@ export const transactionButtonState = (
     execute: TransactionStageStatus,
     wait: TransactionStageStatus,
     write?: () => void,
-    disableOnSuccess: boolean = true,
+    disableOnSuccess: boolean = true
 ) => {
     const loading =
         prepare.status === "loading" ||
@@ -87,6 +87,11 @@ export const ERC20DepositForm: FC<ERC20DepositFormProps> = (props) => {
             }),
         [tokens],
     );
+    );
+};
+
+export const ERC20DepositForm: FC<ERC20DepositFormProps> = (props) => {
+    const { applications, tokens } = props;
 
     const [advanced, { toggle: toggleAdvanced }] = useDisclosure(false);
 
@@ -225,7 +230,7 @@ export const ERC20DepositForm: FC<ERC20DepositFormProps> = (props) => {
             approve,
             approveWait,
             approve.write,
-            false,
+            false
         );
     const { disabled: depositDisabled, loading: depositLoading } =
         transactionButtonState(
@@ -233,7 +238,7 @@ export const ERC20DepositForm: FC<ERC20DepositFormProps> = (props) => {
             deposit,
             depositWait,
             deposit.write,
-            true,
+            true
         );
 
     return (
--- a/packages/ui/src/ERC20DepositForm.tsx
+++ b/packages/ui/src/ERC20DepositForm.tsx
@@ -87,11 +87,6 @@ export const ERC20DepositForm: FC<ERC20DepositFormProps> = (props) => {
             }),
         [tokens],
     );
-    );
-};
-
-export const ERC20DepositForm: FC<ERC20DepositFormProps> = (props) => {
-    const { applications, tokens } = props;
 
     const [advanced, { toggle: toggleAdvanced }] = useDisclosure(false);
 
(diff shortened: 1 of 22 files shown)


packages/ui/src/ERC20DepositForm.tsx
```diff
@@ -70,6 +72,20 @@ export interface ApplicationAutocompleteProps {
     application: string;
     onChange: (application: string) => void;
 }
+
+export const isFirstDeposit = (erc20Address: string, data: string[]) => {
```
nevendyulgerov, line 76: I suggest that we simplify the util in this way:

```
const checkFirstDeposit = (erc20Address: string, data: string[]) =>
    isAddress(erc20Address) &&
    !data.some((contract) => isAddressEqual(contract as Address, erc20Address));
```

This version of the util should work fine for application addresses. For tokens, we can do this:

```
// inside TokenAutocomplete component
const tokenAddresses = useMemo(
    () => tokens.map((t) => t.substring(t.indexOf("0x"))),
    [tokens],
);
...
{checkFirstDeposit(erc20Address, tokenAddresses) && (
    <Alert variant="light" color="yellow" icon={<TbAlertCircle />}>
        This is the first deposit of that token.
    </Alert>
)}
```

Also, you may want to extract `checkFirstDeposit` function in `/src/utils` folder, because we'll probably want to use it in other components like EtherDeposit or RawInput.

brunomenezes:
@dandheedge, Rebase the PR. The #54 was merged.

packages/ui/src/ERC20DepositForm.tsx
brunomenezes: Honestly, guys, I think the ideal solution is a quick ping on the graphQL API for both `Applications` and `Tokens`. I mentioned before that we can provide a pre-filled autocomplete list with 10 ~ 15 items and fire the correct query on input change. That will save us from all the logic involved here—Subsquid `where clause` is quite flexible. 

Here is a reference of one I did for `Dapp Connect` feature, but I do understand they are components living in different packages, and one has direct access to things the other doesn't and that is ok. Because it is just a matter of improving the lifecycle of the component API coming from the `packages/ui`. [code ref](https://github.com/cartesi/rollups-explorer/blob/main/apps/web/src/components/connectionForm.tsx#L110) 


cc: @nevendyulgerov, @dandheedge

dandheedge: In the most recent commit, I've implemented hooks to retrieve applications and tokens using GraphQL. This allows the component to access the necessary data

@brunomenezes @nevendyulgerov

packages/ui/src/ERC20DepositForm.tsx
```diff
@@ -159,6 +164,8 @@ export const ERC20DepositForm: FC<ERC20DepositFormProps> = ({
         }
     }, [erc20.data]);
 
+    useEffect(() => {}, [tokens]);
```
nevendyulgerov, line 167: Empty effect, please remove it if you don't need it.

apps/web/src/components/shell.tsx
```diff
@@ -48,20 +55,20 @@ const Shell: FC<{ children: ReactNode }> = ({ children }) => {
     const { isConnected } = useAccount();
     const { chain } = useNetwork();
     const { colorScheme, toggleColorScheme } = useMantineColorScheme();
-    const [{ data: applicationData }] = useApplicationsQuery({
-        variables: {
-            limit: 1000,
-        },
+    const changeTokenQuery = (address: string) => {
```
nevendyulgerov, line 58: Hey, @dandheedge , this implementation relies on the `Shell` component holding applications and tokens state for the forms. However, with https://github.com/cartesi/rollups-explorer/pull/36/files this will significantly change so that a new `SendTransaction` component controls these states.

Due to this, I recommend that you use https://github.com/cartesi/rollups-explorer/pull/36 as base for your changes here, so that we can sync up on the touching pieces between your branch and `feature/21-create-form-for-sending-raw-input-to-application`. Otherwise, we'll have a bad time resolving conflicts regardless who merges their PR first.

packages/ui/src/ERC20DepositForm.tsx
brunomenezes: @nevendyulgerov, start the review and mark it as `Request changes` for signalling.

brunomenezes (changes requested):
Hey @dandheedge, I tested these changes, and now the copy/paste is working for the application and token address inputs for the ERC-20 deposit. Still, if I select an item in the autocomplete for the token, it displays the warning right away 😕. Have a look at the gif below. 



autocomplete-selection-displays-warning

brunomenezes (approved):

Issue #70, opened 2023-11-21
Age collumn should show the time as an alt text


endersonmaia:
## 📄 Context

When navigating through the inputs list, it would be nice to know the exact time an input happened.

Although the Age field showing the time since the input happens is a good UX, having an option to show the exact time the input was processed is useful for troubleshooting.

![image](https://github.com/cartesi/rollups-explorer/assets/370/47cd8406-0d60-4a36-9c3e-0db26a506696)

## ✔️ Solution

~~One possible solution is to show the date and time as an ALT text on-hover and~~ use the ISO 8061 format

### Update

* Update the `time` column to change based on clicks in the header text. 
 * The header text should switch between `Age` and `Timestamp (UTC)`
 * The header should have a link colour to diff from the others.
 * The cursor should become a 'pointer' when the mouse is over the header text.
 * **Age** content displays using the current format
 * **Timestamp (UTC)** should use the ISO 8061 format.
 
 ## 📈 Subtasks

- [x] Add link colour to the header text of the `Time` column. 
- [x] Hover the `Time` column header text makes the cursor a `pointer`. 
- [x] Click in the `Time` column to make the text switch between `Age` and `Timestamp (UTC)`.
- [x] When **Age** is active, the row content uses the current `time since` format.
- [x] When **Timestamp (UTC)** is active, the row content uses the ISO 8601 format. 
- [x] Add test cases.

 ## 🎯 Definition of Done

- [x] Clicking on the header `Age`, it switches to `Timestamp (UTC)`, and I see the row's content using ISO 8061 format.
- [x] Clicking on the header `Timestamp (UTC)`, it switches to `Age`, and I see the current format (Time since)

brunomenezes:
Hey @endersonmaia, 
That's a good one. I will make some suggestions/adjustments for the `solution` session (I will mark it as **update**). I get what you mean by `Alt text`, but this one is used only on `<img />` or `<figure />` in case of failure and for accessibility too.

tuler:
Another solution used by either etherscan or blockscout is to click on the column header to switch from age to timestamp.

brunomenezes:
> Another solution used by either etherscan or blockscout is to click on the column header to switch from age to timestamp.

That is a good one. Let's settle for that then? @tuler, @endersonmaia

PR #74, opened 2023-11-26
#70 Update time column in inputs table to show date in age or UTC format

nevendyulgerov:


brunomenezes:
Hi @nevendyulgerov,
All the CI and Vercel fail when building the `Web` application and checking the Types.

apps/web/src/components/inputsTable.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <Table>
+            <Table.Thead>
+                <Table.Tr>
+                    <Table.Th>From</Table.Th>
+                    <Table.Th></Table.Th>
+                    <Table.Th>To</Table.Th>
+                    <Table.Th>Method</Table.Th>
+                    <Table.Th>Index</Table.Th>
+                    <Table.Th>
+                        <UnstyledButton
```
brunomenezes, line 28: Hey @nevendyulgerov, 
Why use the `<UnstyleButton />` that leads to defining colours and other styles by hand when you could use it like below?

```jsx
<Button
    variant="transparent"
    onClick={onChangeTimeColumnType}
>
    {timeType === "age" ? "Age" : "Timestamp (UTC)"}
</Button>
```

You get all by default. Overall, we want fewer to no custom-defined styles in the components and more things in the theme. For example, if we change the primary colour in the theme to `teal` for example that would not reflect. I want to avoid the same mistakes made in the Explorer staking.

brunomenezes (changes requested):

brunomenezes (approved):

Issue #91, opened 2023-12-19
Implement not-found for application details page


nevendyulgerov:
## 📄 Context

Currently, when you try to visit an application page for an app that doesn't exist, you see:
![Screenshot 2023-12-19 at 17 52 54](https://github.com/cartesi/rollups-explorer/assets/6005179/b5989a31-f709-4ad4-88b4-fb52e7fe9e65)

The error boundary is being invoked and displayed.

## ✔️ Solution

We can improve this experience and implement [not-found](https://nextjs.org/docs/app/api-reference/file-conventions/not-found) file in combination with using the [notFound](https://nextjs.org/docs/app/api-reference/functions/not-found) function for the `/applications/[address]` page.

Suggested UI:
![Screenshot 2023-12-19 at 18 00 11](https://github.com/cartesi/rollups-explorer/assets/6005179/857b2e5f-f7a5-49d1-a7d2-3f81d71c4b27)

## 🎯 Definition of Done

- [ ] Create a generic "not-found" component that we can use for this case and others.
- [ ] Not found page is presented when you try to open an application page for app that doesn't exist on `/applications/[address]` page.

PR #92, opened 2023-12-21
#91 Add custom not-found UI for root and applications details page

nevendyulgerov:


apps/web/src/app/applications/[address]/page.tsx
```diff
@@ -18,7 +17,36 @@ export async function generateMetadata({
     };
 }
 
-const ApplicationPage: FC<ApplicationPageProps> = ({ params }) => {
+async function getApplication(address: string) {
+    const client = getUrqlClient();
+    const result = await client.query(
+        gql`
```
Comment (brunomenezes) [inline apps/web/src/app/applications/[address]/page.tsx:23]: @nevendyulgerov, you don't need to redefine the GQL for the `application`; we use graphql-generator tools, making queries/types and document definitions available.  So we don't need to repeat ourselves and spread this definition around the code base. So you can use the types and also have auto-completion.


```js
client.query<ApplicationsQuery, ApplicationsQueryVariables>(
        ApplicationsDocument,
        {
            limit: 1,
            where: {
                id_containsInsensitive: address
            }
        }
    )
```
I would change the `id_containsInsensitive` to `id_eq` (that makes a difference in the BE query) since that is supposed to be a full address. I would only apply a `toLowerCase()` for the address value. 


Besides that, I think that `not-found` here is good to have, but it is unlikely someone will type the address in full in the address bar (maybe a copy/paste).

brunomenezes (changes requested):

apps/web/src/app/applications/[address]/page.tsx
Comment (nevendyulgerov) [reply apps/web/src/app/applications/[address]/page.tsx:23]: Hi, @brunomenezes , yes, that's what I initially went for.

However, this didn't work because the autogenerated graphql file contains imports to client-side only libraries like `urql`. When I'm importing from `graphql` folder, like this:

```
import {
    ApplicationsDocument,
    ApplicationsQuery,
    ApplicationsQueryVariables,
} from "../../../graphql";
```

I get the following error:

```
⨯ ../node_modules/urql/dist/urql.es.js (11:8) @ r
⨯ createContext only works in Client Components. Add the "use client" directive at the top of the file to use it. Read more: https://nextjs.org/docs/messages/context-in-server-component

```

I looked into this and the issue is caused by `createContext` react api being used inside some of `urql` exports. Since the `notFound` function can only be triggered server-side, the page needs to be server-side component and also importing from `urql` and thus the generated `graphql` exports is not possible because of that. And due to this I resorted to redefining the graphql for the `application`

Generally, the above ☝️  issue can be mitigated by importing from `urql/core`. That's where `getUrqlClient` imports from:

```
import { Client, cacheExchange, createClient, fetchExchange } from "urql/core";
...
```
This was required, so that there is a unified client for both client and server-side code.

I do agree that this is a rare case that we'll hit (visiting a non-existing application address) but I guess it's nice to also have it covered. However, if you think that the added complexity is not justified, we can ignore it and not implement this altogether.

Comment (brunomenezes) [reply apps/web/src/app/applications/[address]/page.tsx:23]: I see! That is fixed for any rollups-related GraphQL action that we want to do for client/server code, but for the explorer GraphQL, that was postponed just because of priorities. I will push additional changes to the GraphQL codegen here on your branch. Then you can check the graphql folder and try to do the original idea we discussed. The types mentioned here will be inside the `operations.ts`. I will also create an issue and assign you to remove the deprecated config that generates the `index.ts` and refactor all the imports of the hooks used in the web app (The hooks will not be gone; they will be in a separate file; the same way as for the `rollups/).

Comment (brunomenezes) [reply apps/web/src/app/applications/[address]/page.tsx:23]: @nevendyulgerov, 
Also, I would like to mention that if you have a look in the graphQL playground https://sepolia.api.cartesiscan.io/graphql, you will see that it has an `applicationById` query that is exactly what you want for that case. So I think that is better than using the `Applications` one that you have where + limit clause since the ID is known and eliminating the array return since its a `findOne` kind of query.

Comment (nevendyulgerov) [reply apps/web/src/app/applications/[address]/page.tsx]: Hey @brunomenezes , thanks for helping with the graphql definitions! I pushed the changes you requested.

brunomenezes (approved):

Issue #72, opened 2023-11-23
Add ERC-721 Deposit Support


brunomenezes:
## 📄 Context

DApps accept deposits of ERC-721 tokens by interacting with the [ERC721Portal](https://next.docs.cartesi.io/cartesi-rollups/api/json-rpc/portals/ERC721Portal/) smart contract.

Before the transfer, the caller should change the token's approved Address to the **Portal Address**.

## ✔️ Solution

Documentation about the ERC721Portal transfer:
https://next.docs.cartesi.io/cartesi-rollups/api/json-rpc/portals/ERC721Portal/

EIP-721: https://eips.ethereum.org/EIPS/eip-721
EIP-165: https://eips.ethereum.org/EIPS/eip-165

### Overall Form Structure
* The input for the application should use the autocomplete by fetching the Apps list from the API. 
* The input for ERC-721 contract. _ps: Autocomplete will come in the future (new feature in the API required)_
* The input for the token-id. **(A bit more details below)**
* There are two new fields for the `Advanced` area, unlike the ERC-20, which has one.
  * One for the Base layer.
  * One for the execution layer.

For the token-id input
* Below the input, display the balance (e.g. `balance: 2 BAYC` for Bored ape); if the symbol name is not available, only display the numeric part.
* When the **enumerable extension** `tokenOfOwnerByIndex` is not available
  * The input should be `text` type for the user to manually enter the token-id.  
* When the **enumerable extension** `tokenOfOwnerByIndex` is available
  * If the user owns 1+ tokens, the input should be a `select` listing all token IDs owned.   
  * If the user owns nothing, just fallback to the `text` type described above.

### Extra
Some mainnet ERC-721 contracts to help. 

Bored Ape Yacht Club: 0xBC4CA0EdA7647A8aB7C2061c2E118A18a936f13D
CryptoKitties: 0x06012c8cf97BEaD5deAe237070F9587f8E7A266d (Older does not have enumerable extension)
  
## 📈 Subtasks

- [x] Create the form to deposit ERC-721 tokens.
- [x] Implement the `approve` transaction.
- [x] Implement the `depositERC721Token` transaction
- [x] Add test cases

## 🎯 Definition of Done

- [x] The transaction can be completed with success.

brunomenezes:
@tuler, is there anything I missed in the details of this issue? 

[...]

PR #96, opened 2024-01-08
#72 Add ERC-721 Deposit form

nevendyulgerov:


brunomenezes (changes requested):
Hi @nevendyulgerov, 
I got a few things. I will describe the situation and include gifs and stack trace where applicable. 

1. The first one is that when I use an ERC-721 contract with the enumerable extension, I can see my token IDs as expected, i.e. [5, 6]. I select the token ID six and then switch to another NFT contract that does not have this extension implemented. The input is filled with the previous value, in that case, 6, even though the only token I had for that contract was ID 4. Hence, a clean-up when switching the ERC-721 contract address would be better than displaying an unrelated value. 
2. The second situation is after a deposit transaction is confirmed, I believe the state of the form is kept the same instead of a refresh or maybe a clean-up; for that case, I deposited my token-id five, and after the confirmation message, the same id was still on screen, I clean-up the contract address and put it back and the balance seems to be updated. However, the ID five and confirmation message were still on the screen, and after a bit, an error message was splitting the screen; I have the impression an auto deposit fired up when I put the contract address back and failed since I did not have the token ID 5 anymore. For that one, a full refresh and reset in the token-id form value will probably avoid this weird state and the deposit action firing up without the user clicking the button. 


### First case 

token-id-not-cleaned-after-address-change


### Second case 

error-after-tx_confirmation-no-clean-up


<details>
<summary><h3> Stacktrace </h3></summary>

```log
The contract function "depositERC721Token" reverted with the following signature:
0x177e802f

Unable to decode signature "0x177e802f" as it was not found on the provided ABI.
Make sure you are using the correct ABI and that the error exists on it.
You can look up the decoded signature here: https://openchain.xyz/signatures?query=0x177e802f.
 
Contract Call:
  address:   0x237F8DD094C0e47f4236f12b4Fa01d6Dae89fb87
  function:  depositERC721Token(address _token, address _dapp, uint256 _tokenId, bytes _baseLayerData, bytes _execLayerData)
  args:                        (0x569DABb4F67770cc094d09Fe4bf4202557d2f456, 0x4cA2f6935200b9a782A78f408F640F17B29809d8, 5, 0x3078, 0x3078)
  sender:    0xa074683B5BE015F053b5Dceb064C41fC9D11B6E5

Docs: https://viem.sh/docs/contract/decodeErrorResult.html
Version: viem@1.19.10
```

</details>

brunomenezes (changes requested):
@nevendyulgerov, 
I see the form is cleaned, but after the deposit, After adding the app and erc-721 addresses, the moment I type the token ID, it shows error messages even though I am the owner of the token ID in the screenshot and it is supposed to be in a "clean" state. Something is not fresh with the Wagmi hooks and values passed down. 

image

nevendyulgerov:
> @nevendyulgerov, I see the form is cleaned, but after the deposit, After adding the app and erc-721 addresses, the moment I type the token ID, it shows error messages even though I am the owner of the token ID in the screenshot and it is supposed to be in a "clean" state. Something is not fresh with the Wagmi hooks and values passed down.
> 
> PS: There is a small rebase to be done.
> 
> image

Hey @brunomenezes , thanks for the feedback. I fixed that issue by resetting the approve and deposit hooks after successful deposit. I added a 3 seconds delay for that reset so that the successful deposit alert is visible for some time on the screen.

brunomenezes:
> Hey @brunomenezes , thanks for the feedback. I fixed that issue by resetting the approve and deposit hooks after successful deposit. I added a 3 seconds delay for that reset so that the successful deposit alert `Transaction confirmed` is visible for some time on the screen.

@nevendyulgerov, I see. I wonder if it would be good to keep the `transition` state with the feedback as is in the form (waiting tx, etc.), but the successful one being dispatched to the `Notification` system available (Like the Connections feature does), so we decouple the success feedback from the hooks state since we want to apply a reset. I am not a fan of `timeouts.`. I did a roll of tests and sent you a message on Discord.

nevendyulgerov:
> > Hey @brunomenezes , thanks for the feedback. I fixed that issue by resetting the approve and deposit hooks after successful deposit. I added a 3 seconds delay for that reset so that the successful deposit alert `Transaction confirmed` is visible for some time on the screen.
> 
> @nevendyulgerov, I see. I wonder if it would be good to keep the `transition` state with the feedback as is in the form (waiting tx, etc.), but the successful one being dispatched to the `Notification` system available (Like the Connections feature does), so we decouple the success feedback from the hooks state since we want to apply a reset. I am not a fan of `timeouts.`. I did a roll of tests and sent you a message on Discord.

Hi, @brunomenezes , I decoupled the transaction alert from the deposit success message so that on successful deposit a mantine alert is displayed.

brunomenezes (approved):

PR #114, opened 2024-02-08
#111 Add search UI

### LINKED ISSUE #111: Add Search UI
## 📄 Context

With the increasing number of data, users often find it challenging to quickly locate specific data for input or application.

## ✔️ Solution

Develop a search UI component that allows users to efficiently filter and find the data they are looking for. The search component would be on two pages.

## 📈 Subtasks

- [ ] Implement a search bar that enables users to effectively filter and find specific data on `Applications` page. 
- [ ] Implement a search bar that enables users to effectively filter and find specific data on `Inputs` page. 

## 🎯 Definition of Done

- [ ]  A fully operational search bar that accurately filters and displays the specific data based on user input.

### Example 

Screenshot 2024-01-30 at 13 17 06

dandheedge:
Close #111

### DIFF
--- a/apps/web/src/components/inputs/inputs.tsx
+++ b/apps/web/src/components/inputs/inputs.tsx
@@ -2,10 +2,11 @@
 
 import { Stack } from "@mantine/core";
 import { FC, useCallback, useState } from "react";
-import { InputOrderByInput } from "../../graphql/explorer/types";
-import { useInputsQuery } from "../../graphql/explorer/hooks/queries";
+import { InputOrderByInput, useInputsConnectionQuery } from "../../graphql";
+import { checkQuery } from "../../lib/query";
 import InputsTable from "../inputs/inputsTable";
 import Paginated from "../paginated";
+import Search from "../search";
 
 export type InputsProps = {
     orderBy?: InputOrderByInput;
@@ -16,15 +17,20 @@ const Inputs: FC<InputsProps> = ({
     orderBy = InputOrderByInput.TimestampDesc,
     applicationId,
 }) => {
+    const [query, setQuery] = useState("");
     const [limit, setLimit] = useState(10);
     const [page, setPage] = useState(1);
     const after = page === 1 ? undefined : ((page - 1) * limit).toString();
-    const [{ data, fetching }] = useInputsQuery({
+
+    const [{ data: data, fetching: fetching }] = useInputsConnectionQuery({
         variables: {
             orderBy,
-            applicationId: applicationId?.toLowerCase(),
             limit,
             after,
+            where: checkQuery(
+                query.toLowerCase(),
+                applicationId?.toLowerCase(),
+            ),
         },
     });
     const inputs = data?.inputsConnection.edges.map((edge) => edge.node) ?? [];
@@ -34,8 +40,13 @@ const Inputs: FC<InputsProps> = ({
         setPage(page);
     }, []);
 
+    const onSubmitQuery = useCallback((query: string) => {
+        setQuery(query);
+    }, []);
+
     return (
         <Stack>
+            <Search onSubmit={onSubmitQuery} />
             <Paginated
                 fetching={fetching}
                 totalCount={data?.inputsConnection.totalCount}
--- a/apps/web/src/components/inputs/inputs.tsx
+++ b/apps/web/src/components/inputs/inputs.tsx
@@ -2,7 +2,8 @@
 
 import { Stack } from "@mantine/core";
 import { FC, useCallback, useState } from "react";
-import { InputOrderByInput, useInputsConnectionQuery } from "../../graphql";
+import { useInputsConnectionQuery } from "../../graphql/explorer/hooks/queries";
+import { InputOrderByInput } from "../../graphql/explorer/types";
 import { checkQuery } from "../../lib/query";
 import InputsTable from "../inputs/inputsTable";
 import Paginated from "../paginated";
--- a/apps/web/src/components/inputs/inputs.tsx
+++ b/apps/web/src/components/inputs/inputs.tsx
@@ -41,13 +41,9 @@ const Inputs: FC<InputsProps> = ({
         setPage(page);
     }, []);
 
-    const onSubmitQuery = useCallback((query: string) => {
-        setQuery(query);
-    }, []);
-
     return (
         <Stack>
-            <Search onSubmit={onSubmitQuery} />
+            <Search isLoading={fetching} onChange={setQuery} />
             <Paginated
                 fetching={fetching}
                 totalCount={data?.inputsConnection.totalCount}
--- a/apps/web/src/components/inputs/inputs.tsx
+++ b/apps/web/src/components/inputs/inputs.tsx
@@ -1,6 +1,7 @@
 "use client";
 
 import { Stack } from "@mantine/core";
+import { useDebouncedValue } from "@mantine/hooks";
 import { FC, useCallback, useState } from "react";
 import { useInputsConnectionQuery } from "../../graphql/explorer/hooks/queries";
 import { InputOrderByInput } from "../../graphql/explorer/types";
@@ -23,13 +24,15 @@ const Inputs: FC<InputsProps> = ({
     const [page, setPage] = useState(1);
     const after = page === 1 ? undefined : ((page - 1) * limit).toString();
 
+    const [queryDebounced] = useDebouncedValue(query, 500);
+
     const [{ data: data, fetching: fetching }] = useInputsConnectionQuery({
         variables: {
             orderBy,
             limit,
             after,
             where: checkQuery(
-                query.toLowerCase(),
+                queryDebounced.toLowerCase(),
                 applicationId?.toLowerCase(),
             ),
         },
--- a/apps/web/src/components/search.tsx
+++ b/apps/web/src/components/search.tsx
@@ -1,21 +1,27 @@
 import { Box, Loader, TextInput } from "@mantine/core";
-import { useDebouncedState } from "@mantine/hooks";
-import React, { useEffect } from "react";
+import React, { useCallback, useEffect, useState } from "react";
 import { CiSearch } from "react-icons/ci";
 import { useQueryParams } from "../hooks/useQueryParams";
 export type SearchProps = {
-    isLoading: Boolean;
+    isLoading: boolean;
     onChange: (query: string) => void;
 };
 
 const Search: React.FC<SearchProps> = ({ onChange, isLoading }) => {
     const { query, updateQueryParams } = useQueryParams();
-    const [keyword, setKeyword] = useDebouncedState("", 500);
+    const [keyword, setKeyword] = useState<string>("");
 
     useEffect(() => {
         updateQueryParams(keyword);
         onChange(keyword);
-    }, [query, onChange, keyword]);
+    }, [query, onChange, keyword, updateQueryParams]);
+
+    const onSearch = useCallback(
+        (event: React.ChangeEvent<HTMLInputElement>) => {
+            setKeyword(event.target.value);
+        },
+        [],
+    );
 
     return (
         <Box w={{ sm: "10%%", lg: "50%" }} mb={{ sm: "1rem", lg: "-3.25rem" }}>
@@ -30,10 +36,8 @@ const Search: React.FC<SearchProps> = ({ onChange, isLoading }) => {
                 }
                 size="md"
                 data-testid="search-input"
-                defaultValue={keyword}
-                onChange={(event) => {
-                    setKeyword(event.target.value);
-                }}
+                value={keyword}
+                onChange={onSearch}
             />
         </Box>
     );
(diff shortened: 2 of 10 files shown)


apps/web/src/components/search.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { useQueryParams } from "../hooks/useQueryParams";
+
+export type SearchProps = {
+    onSubmit: (query: string) => void;
+};
+
+const Search: React.FC<SearchProps> = ({ onSubmit }) => {
+    const { query, updateQueryParams } = useQueryParams();
+
+    const form = useForm({
```
nevendyulgerov, line 13: I think we can go with simple react state here. The reason for this is that we are not using much from the `useForm` hook other than the `initialValues` field, like validation.

apps/web/src/components/search.tsx
```diff
@@ -0,0 +1,54 @@
...
+        initialValues: {
+            query: "",
+        },
+    });
+    const inputRef = useRef<HTMLInputElement>(null);
+
+    const handleKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
```
nevendyulgerov, line 20: This is nice way to perform the search but maybe we can use debouncing and just search after the user has stopped entering symbols in the input? I think some users will expect the input to automatically perform a search once they stopped typing.

apps/web/src/components/search.tsx
```diff
@@ -0,0 +1,54 @@
...
+        if (event.key === "Enter") {
+            updateQueryParams(form.values.query);
+            form.onSubmit((values) => onSubmit(values.query))();
+            if (inputRef.current) {
+                inputRef.current.blur();
```
nevendyulgerov, line 25: Not sure if we really want to blur the input. What if the user wants to continue changing the query? They would have to focus again the input to be able to do that.

apps/web/src/components/search.tsx
```diff
@@ -0,0 +1,54 @@
...
+            }
+        }
+    };
+
+    useEffect(() => {
+        onSubmit(query);
+    }, [query, onSubmit]);
+
+    return (
+        <>
```
nevendyulgerov, line 35: We can remove this fragment and just use the `Box` as a wrapper. Less nesting this way.

nevendyulgerov (commented):
Nice work, @dandheedge ! I think we can make some slight adjustments to make the search a bit more intuitive.

apps/web/src/components/search.tsx
```diff
@@ -0,0 +1,54 @@
...
+            <Box
+                w={{ sm: "10%%", lg: "50%" }}
+                mb={{ sm: "1rem", lg: "-3.25rem" }}
+            >
+                <TextInput
```
nevendyulgerov, line 40: Maybe we can add a spinner using the `leftSection` prop to notify users that we are currently searching for their query. This will improve the UX.

apps/web/src/components/search.tsx
```diff
@@ -0,0 +1,54 @@
...
+                    placeholder="Search by Address / Txn Hash / Index"
+                    {...form.getInputProps("query")}
+                    onKeyDown={handleKeyDown}
+                    ref={inputRef}
+                    error={form.errors.query}
```
nevendyulgerov, line 45: Do we have errors for this field? If not, let's remove the error prop.

apps/web/src/lib/query.ts
```diff
@@ -0,0 +1,58 @@
...
+};
+
+/**
+ *
+ * @param input
+ * @param applicationId
+ * @returns
+ */
+
+export const checkQuery = (
```
nevendyulgerov, line 28: Let's please cover this utility with some unit tests to be on the safe side.

apps/web/test/components/search.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+            query: "",
+            updateQueryParams: (value: string) => undefined,
+        });
+    });
+    afterEach(() => {
+        vi.clearAllMocks();
+        cleanup();
+    });
+
+    it("should display search input element", async () => {
```
nevendyulgerov, line 48: Do we perform any async operation in those unit tests? If not, we can make them synchronous.

apps/web/src/components/search.tsx
```diff
@@ -0,0 +1,54 @@
...
+                    size="md"
+                    data-testid="search-input"
+                />
+            </Box>
+        </>
+    );
+};
+
+export default Search;
```
nevendyulgerov, line 54: While testing, I noticed a slight glitch in the UI when there are no results found:

https://github.com/cartesi/rollups-explorer/assets/6005179/1b5741f9-8a58-4887-ab0e-d4bbf72e257f

apps/web/src/components/inputs/inputs.tsx
```diff
@@ -34,8 +41,13 @@ const Inputs: FC<InputsProps> = ({
         setPage(page);
     }, []);
 
+    const onSubmitQuery = useCallback((query: string) => {
+        setQuery(query);
+    }, []);
+
     return (
         <Stack>
+            <Search onSubmit={onSubmitQuery} />
```
nevendyulgerov, line 50: We can do this instead:

```
<Search onSubmit={setQuery} />
```

And get rid of the cached `onSubmitQuery` function. The setter coming from `useState` has stable reference.

apps/web/src/hooks/useQueryParams.tsx
```diff
@@ -0,0 +1,28 @@
+import { usePathname, useRouter, useSearchParams } from "next/navigation";
+import { useCallback } from "react";
+
+export type UseQueryParamsReturn = {
+    query: string;
+    updateQueryParams: (value: string) => void;
+};
+
+export const useQueryParams = (): UseQueryParamsReturn => {
```
nevendyulgerov, line 9: Can we cover this hook with some unit tests please?

apps/web/src/components/search.tsx
dandheedge: @nevendyulgerov thank you for your input. Regarding performance, what is your opinion on the optimal debounce timeout? Should we align it with the standard `400ms` used in other forms, or would a longer duration be more appropriate

nevendyulgerov: Yes, I think we can start with `400ms` and see how it goes. @brunomenezes what do you think? Do we need a longer delay here?

brunomenezes: I think around 400 ~ 500 ms is a good starting point. @dandheedge you can check a reference [here](https://github.com/cartesi/rollups-explorer/blob/main/apps/web/src/components/connection/connectionForm.tsx#L173)

apps/web/src/components/search.tsx
```diff
@@ -0,0 +1,42 @@
...
+};
+
+const Search: React.FC<SearchProps> = ({ onChange, isLoading }) => {
+    const { query, updateQueryParams } = useQueryParams();
+    const [keyword, setKeyword] = useDebouncedState("", 500);
+
+    useEffect(() => {
+        updateQueryParams(keyword);
+        onChange(keyword);
+    }, [query, onChange, keyword]);
```
nevendyulgerov, line 18: Missing dependency here - `updateQueryParams`.

apps/web/src/components/search.tsx
```diff
@@ -0,0 +1,42 @@
+import { Box, Loader, TextInput } from "@mantine/core";
+import { useDebouncedState } from "@mantine/hooks";
+import React, { useEffect } from "react";
+import { CiSearch } from "react-icons/ci";
+import { useQueryParams } from "../hooks/useQueryParams";
+export type SearchProps = {
+    isLoading: Boolean;
```
nevendyulgerov, line 7: `Boolean` is an object type. Please use the primitive type `boolean` here the instead.

apps/web/src/components/search.tsx
```diff
@@ -0,0 +1,42 @@
...
+                rightSection={
+                    keyword &&
+                    isLoading && (
+                        <Loader size={"xs"} aria-label="loader-input" />
+                    )
+                }
+                size="md"
+                data-testid="search-input"
+                defaultValue={keyword}
+                onChange={(event) => {
```
nevendyulgerov, line 34: Just a suggestion, please consider applying a `useCallback` here to have a stable functional reference to the `onChange` callback:

```
const onChange = useCallback((event) => {
  setKeyword(event.target.value);
}, []);
...

<TextInput
  ...
  onChange={onChange}
/>
```

apps/web/src/components/search.tsx
```diff
@@ -0,0 +1,42 @@
...
+                leftSection={<CiSearch />}
+                rightSection={
+                    keyword &&
+                    isLoading && (
+                        <Loader size={"xs"} aria-label="loader-input" />
+                    )
+                }
+                size="md"
+                data-testid="search-input"
+                defaultValue={keyword}
```
nevendyulgerov, line 33: I think we should use here `value` instead of `defaultValue` since this is a controlled input.

apps/web/src/hooks/usePaginationParams.tsx
```diff
@@ -19,25 +19,24 @@ export const usePaginationParams = (): UsePaginationReturn => {
...
                 pg: page.toString(),
                 lt: limit.toString(),
             });
-
             router.push(`${pathName}?${urlSearchParams.toString()}`, {
                 scroll: false,
             });
         },
-        [router, pathName],
+        [router, pathName, searchParams],
```
nevendyulgerov, line 39: Missing dependency `query`, please add it to the dependencies array.

apps/web/src/hooks/useQueryParams.tsx
```diff
@@ -0,0 +1,33 @@
...
+            }
+            router.push(`${pathName}?${urlSearchParams.toString()}`, {
+                scroll: false,
+            });
+        },
+        [router, pathName, searchParams],
+    );
+    return useMemo(
+        () => ({ query, updateQueryParams }),
+        [query, updateQueryParams, searchParams],
```
nevendyulgerov, line 31: Unnecessary dependency here `searchParams`, please remove it from the dependencies array.

apps/web/test/components/search.test.tsx
```diff
@@ -0,0 +1,97 @@
...
```
nevendyulgerov, line 56: Not sure that we need this test. It just verifies that react renders the component correctly. I think we can rely on react to work as expected in this case.

apps/web/test/hooks/useQueryParams.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+        usePathname: vi.fn(() => "/"),
+        useRouter: () => ({
+            push: vi.fn(),
+        }),
+        useSearchParams: vi.fn(() => new URLSearchParams()),
+    };
+});
+
+const queryAddress = "0xdd1f9B83507327f29C2C1Bb42011faD5fb482dc6";
+const mockedUsePathname = vi.mocked(usePathname);
```
nevendyulgerov, line 24: `mockedUsePathname` and `mockedUseRouter` are unused. Please remove those two if we don't need them.

nevendyulgerov (commented):
Thanks for making those tweaks, Dandhi! I left a couple more comments, small stuff like missing or unneeded dependencies for hooks.

nevendyulgerov (approved):
LGTM! I made a couple of small tweaks related to `usePaginationParams` and `useQueryParams` custom hooks. I hope you don't mind @dandheedge :)

apps/web/src/hooks/usePaginationParams.tsx
```diff
@@ -38,5 +39,8 @@ export const usePaginationParams = (): UsePaginationReturn => {
         },
         [query, router, pathName],
     );
-    return [{ page, limit }, updateParams];
+    return useMemo(
```
brunomenezes, line 42: Hey @nevendyulgerov, Do you see a necessity for that `useMemo` here? because there is no expensive calculation happening here; it is just the returned result of the above operations. May be even "slower" than just returning the results since an evaluation needs to occur under the hood.

nevendyulgerov: Hey @brunomenezes , this was mostly a suggestion so that the returned array `[{ page, limit }, updateParams]` has a stable reference. Otherwise it will have a new reference (in memory) in each pass, which may result in needless updates to cached callbacks or effects for which any of the values in the array are being used as dependencies.

brunomenezes: Hey @nevendyulgerov, I see. But the returned array/tuple is a standard react hook return, right? Typically, you do a destructuring on the left-hand side, so the array/tuple itself does not matter. Now, the `updateParams` was already using a `useCallback`, so it is not a new function on every render (so that is covered for the case above); the first item in the array or `data` would be a problem if you decide to throw the object in the array-deps, but usually you will also do destructuring on the values that matter. These values are primitives and react uses the `Object.is()` under the hood to check changes, therefore not causing any unnecessary triggers on subsequent `array-deps`. 

There are valid cases for adding an entire object to such a deps array, but it comes with some pitfalls.

My point was more related to avoiding **premature optimization** due to the points I mentioned.

Issue #101, opened 2024-01-15
Add voucher execution


brunomenezes:
## 📄 Context

Be able to execute a voucher generated by a Rollups Application.

## ✔️ Solution

When a connection exists and a voucher is available in the `input details` area, Display a button to execute the voucher **if it has not been executed yet**. The arguments for execution can be found in the `Voucher` type returned by the rollup graphQL. 

<details>
<summary>GraphQL references</summary>

HoneyPot graphQL playground: https://honeypot.sepolia.rollups.staging.cartesi.io/graphql

```graphql
type Voucher {
  index: Int!
  input: Input!
  destination: String!
  payload: String!
  proof: Proof
}
```
</details>

<details>
<summary>CartesiDApp contract references</summary>
<br>
Methods:

* [wasVoucherExecuted](https://github.com/cartesi/rollups-contracts/blob/77ff2c08928b871348b6e2bf3b6f5d7f3c95c9dc/onchain/rollups/contracts/dapp/CartesiDApp.sol#L148)
* [executeVoucher](https://github.com/cartesi/rollups-contracts/blob/77ff2c08928b871348b6e2bf3b6f5d7f3c95c9dc/onchain/rollups/contracts/dapp/CartesiDApp.sol#L103C14-L103C28)
</details>

Wagmi reference: [Wagmi-cli docs](https://wagmi.sh/cli/getting-started)

## 📈 Subtasks

- [x] Update rollups-wagmi package to include CartesiDApp ABI for code generation. 
- [x] In the input details area, only the voucher area should present the button for action.
- [x] A spinner should be displayed for feedback While checking if the voucher was executed.
- [x] When the voucher was executed, the button should be disabled, and the copy displayed is `Executed`.
- [x] Use [useWaitForTransaction](https://1.x.wagmi.sh/react/hooks/useWaitForTransaction) for the transaction loading.
- [x] Display a `Execute` button when the voucher is not executed.
- [x] Add stories to cover the new UI and states (checking / while executing) 
- [x] Add test cases.

## 🎯 Definition of Done

- [x] Be able to execute an Application voucher.

brunomenezes:
@tuler, adding you for review as we discussed.

PR #115, opened 2024-02-12
#101 Add voucher execution

nevendyulgerov:
I added the voucher execution logic in a component called `VoucherExecution` located in `apps/web`. For testing of this logic I used the following [dapp](https://github.com/WeAreProductive/drawing-DApp/) which comes will functionality for adding inputs with vouchers.

### DIFF
--- a/apps/web/src/components/inputs/voucherExecution.tsx
+++ b/apps/web/src/components/inputs/voucherExecution.tsx
@@ -34,12 +34,11 @@ const typeCastProof = (voucher: Partial<Voucher>) => ({
 
 export interface VoucherExecutionType {
     appId: Address;
-    vouchers: Partial<Voucher>[];
+    voucher: Partial<Voucher>;
 }
 
 const VoucherExecution: FC<VoucherExecutionType> = (props) => {
-    const { appId, vouchers } = props;
-    const [voucher] = vouchers;
+    const { appId, voucher } = props;
     const hasVoucherProof =
         typeof voucher?.proof === "object" && voucher?.proof !== null;
 
--- a/apps/web/src/components/inputs/voucherExecution.tsx
+++ b/apps/web/src/components/inputs/voucherExecution.tsx
@@ -71,7 +71,7 @@ const VoucherExecution: FC<VoucherExecutionType> = (props) => {
                 </Flex>
             ) : (
                 <Tooltip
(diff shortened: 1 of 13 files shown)


apps/web/src/components/inputs/voucherExecution.tsx
```diff
@@ -0,0 +1,111 @@
...
+    });
+    const execute = useCartesiDAppExecuteVoucher({
+        args: [
+            voucher.destination as `0x${string}`,
+            voucher.payload as `0x${string}`,
+            typeCastProof(voucher),
+        ],
+        address: appId,
+    });
+    const isExecuted = wasExecuted.data || execute.status === "success";
```
brunomenezes, line 62: Hi @nevendyulgerov, There is one thing that I got once, but I could not reproduce it again in devnet even after increasing the block production time, but I am sure it will happen. When executing the voucher, it instantly shows the "success" message; looking at the code here, it is not waiting for any transaction confirmation. Given the testnet and mainnet latencies, the following situation will likely happen.

1. I clicked to execute the voucher and confirmed the transaction with Metamask.
2. Success message instantly displayed and the button was `disabled` with correct wording.
3. I quickly refresh the screen to check the voucher tab.
4. The `Execute` button was enabled again.
5. Then, after a bit, Metamask displayed the confirmation message.
6. I refreshed again, and the word and state of the button were correct i.e. `Executed` and `Disabled`

So I think an [useWaitForTransaction](https://1.x.wagmi.sh/react/migration-guide#usewaitfortransaction) is necessary like we do on ERC-20 forms as an example [here](https://github.com/cartesi/rollups-explorer/blob/main/packages/ui/src/ERC20DepositForm.tsx#L194), also handling the state of the returned result like [here](https://github.com/cartesi/rollups-explorer/blob/main/packages/ui/src/ERC20DepositForm.tsx#L232) to decide how long the `loader` should stay and avoid any possibility of double execution (in that case would only mean double eth spending, but the end result would be the same).

brunomenezes (changes requested):

brunomenezes (approved):
@nevendyulgerov, it looks good to me. I have a mixed feeling regarding the `<VoucherExecution/>` being located in the apps/web instead of being part of the `packages/ui`. But that is not a merge blocker. Let's discuss this later.

Issue #119, opened 2024-02-15
Warning for not deployed application is presented while loading applications


nevendyulgerov:
## 🙂 Expected behavior

Warning for not deployed application shouldn't appear while loading applications, after a validation error for the applications field is cleared.

## 🫠 Actual behavior

Warning for not deployed application appears while loading applications, after a validation error for the applications field is cleared.

## 🧪 Minimal test case

1. Click Send Transaction
2. Select ERC20 Deposit
3. Enter for application `0x674cb14ea404348b3b87c6e629f862014a72b6e9dasdasd` to cause a validation error for that field
4. Enter correct and existing application address like `0x674cb14ea404348b3b87c6e629f862014a72b6e9`

## 🌎 Environment

- Sepolia network
- Any rollups version

## Video demonstrating the issue:

https://github.com/cartesi/rollups-explorer/assets/6005179/6889ad0b-c6c2-452b-a81b-6c24a46662bf

PR #124, opened 2024-02-27
#119 Tweak warning for undeployed application

nevendyulgerov:
I tweaked the behaviour of applications autocomplete field so that the check for undeployed application is evaluated 1 second after the latest batch of applications is pulled. This way the related warning won't appear while the applications query is executed in the background.

To achieve this, I added a custom hook called `useUndeployedApplication`.

### DIFF
--- a/packages/ui/src/hooks/useUndeployedApplication.ts
+++ b/packages/ui/src/hooks/useUndeployedApplication.ts
@@ -1,0 +1,26 @@
+import { useEffect, useState } from "react";
+import { Address, isAddress, zeroAddress } from "viem";
+
+export default function useUndeployedApplication(
+    address: Address,
+    applications: string[],
+    delay = 1000,
+) {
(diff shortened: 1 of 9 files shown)


packages/ui/src/hooks/useUndeployedApplication.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export default function useUndeployedApplication(
+    address: Address,
+    applications: string[],
+    delay = 1000,
+) {
+    const [isUndeployedApp, setUndeployedApp] = useState(false);
+
+    useEffect(() => {
+        const timeout = setTimeout(() => {
```
dandheedge, line 12: @nevendyulgerov can we use the debounced hooks rathet than write the timeout methods from scratch?

dandheedge (dismissed):

dandheedge (approved):

PR #125, opened 2024-03-05
116. Implement application overview card and table

### LINKED ISSUE #116: Implement Application Overview Card and Table
## 📄 Context

Applications page is not displaying the total number of applications nor does it provide a table listing the applications owned by the user.

## ✔️ Solution

1. Create a summary card in the `/home` page that shows total applications owned by user.
2. Create a table component in the `/applications` that shows all applications owned by user.

**Inspiration**

Screenshot 2024-02-13 at 13 54 29

## 📈 Subtasks

- [x] Create a card that shows total user's applications on the homepage
- [x] Create a table that shows only applications owned by user
- [x] Add test cases for all added logic and UI
- [x] Add stories for the UI under the `workshop` project

## 🎯 Definition of Done

- [x] A card component titled My Component accurately displays the total number of applications the user has.
- [x] Table component that only display the user's Applications
- [x] Correctly use the query on each component
- [x] Existing and new unit tests pass successfully

dandheedge:
Close #116

### DIFF
--- a/apps/web/graphql/queries.graphql
+++ b/apps/web/graphql/queries.graphql
@@ -65,6 +65,12 @@ query stats($applicationId: String) {
     }
 }
 
+query statsApplicationsOwner($applicationId: String) {
+    applicationsConnection(orderBy: id_ASC, where: {owner_startsWith: $applicationId}) {
+        totalCount
+    }
+}
+
 query applications($limit: Int, $where: ApplicationWhereInput) {
     applications(where: $where, limit: $limit) {
         id
--- a/apps/web/graphql/queries.graphql
+++ b/apps/web/graphql/queries.graphql
@@ -65,8 +65,8 @@ query stats($applicationId: String) {
     }
 }
 
-query statsApplicationsOwner($applicationId: String) {
-    applicationsConnection(orderBy: id_ASC, where: {owner_startsWith: $applicationId}) {
+query statsApplicationsOwner($ownerAddress: String) {
+    applicationsConnection(orderBy: id_ASC, where: {owner_eq: $ownerAddress}) {
         totalCount
     }
 }
@@ -112,7 +112,7 @@ query applicationsConnection($orderBy: [ApplicationOrderByInput!]!, $limit: Int,
 }
 
 query applicationsConnectionOwner($orderBy: [ApplicationOrderByInput!]!, $limit: Int, $after: String, $ownerId: String) {
-    applicationsConnection(orderBy: $orderBy, first: $limit, after: $after, where: {owner_startsWith: $ownerId}) {
+    applicationsConnection(orderBy: $orderBy, first: $limit, after: $after, where: {owner_eq: $ownerId}) {
         totalCount
         pageInfo {
             hasNextPage
--- a/apps/web/graphql/queries.graphql
+++ b/apps/web/graphql/queries.graphql
@@ -111,6 +111,23 @@ query applicationsConnection($orderBy: [ApplicationOrderByInput!]!, $limit: Int,
     }
 }
 
+query applicationsConnectionOwner($orderBy: [ApplicationOrderByInput!]!, $limit: Int, $after: String, $ownerId: String) {
+    applicationsConnection(orderBy: $orderBy, first: $limit, after: $after, where: {owner_startsWith: $ownerId}) {
+        totalCount
+        pageInfo {
+            hasNextPage
+            hasPreviousPage
+            startCursor
+            endCursor
+        }
+        edges {
+            node {
+                ...ApplicationItem
+            }
+        }
+    }
+}
+
 query applicationById($id: String!) {
     applicationById(id: $id) {
         ...ApplicationItem
--- a/apps/web/src/components/applications/applications.tsx
+++ b/apps/web/src/components/applications/applications.tsx
@@ -20,7 +20,7 @@ const UserApplications: FC = () => {
             orderBy: ApplicationOrderByInput.IdAsc,
             limit: 20,
             after: "1",
-            ownerId: address,
+            ownerId: address?.toLowerCase(),
         },
     });
     const applications =
--- a/apps/web/src/components/applications/applications.tsx
+++ b/apps/web/src/components/applications/applications.tsx
@@ -1,6 +1,14 @@
 "use client";
 
-import { Stack, Title, useMantineColorScheme } from "@mantine/core";
+import {
+    Group,
+    Pagination,
+    Select,
+    Stack,
+    Text,
+    Title,
+    useMantineColorScheme,
+} from "@mantine/core";
 import { FC, useCallback, useState } from "react";
 import { useAccount } from "wagmi";
 import {
@@ -9,7 +17,7 @@ import {
 } from "../../graphql/explorer/hooks/queries";
 import { ApplicationOrderByInput } from "../../graphql/explorer/types";
 import ApplicationsTable from "../applications/applicationsTable";
-import Paginated from "../paginated";
+import Paginated, { perPageList } from "../paginated";
 import UserApplicationsTable from "./userApplicationsTable";
 
 type StackApplicationProps = {
@@ -20,6 +28,7 @@ type StackApplicationProps = {
  *
  * @returns Component to wrap Application or UserApplication coponent
  */
+
 const StackApplication: FC<StackApplicationProps> = ({ children }) => {
     const { colorScheme } = useMantineColorScheme();
     return (
@@ -40,21 +49,43 @@ const StackApplication: FC<StackApplicationProps> = ({ children }) => {
 
 const UserApplications: FC = () => {
     const { address, isConnected } = useAccount();
+    const [page, setPage] = useState(1);
+    const [limit, setLimit] = useState<number>(10);
+    const after = page === 1 ? undefined : ((page - 1) * limit).toString();
     const [query] = useApplicationsConnectionOwnerQuery({
         variables: {
+            after,
+            limit,
             orderBy: ApplicationOrderByInput.IdAsc,
-            ownerId: address?.toLowerCase(),
+            ownerId: "0xf05d57a5bed2d1b529c56001fc5810cc9afc0335".toLowerCase(),
         },
         pause: !isConnected,
     });
+    const totalCount = query.data?.applicationsConnection.totalCount;
+    const totalPages = Math.ceil(
+        totalCount === undefined || totalCount === 0 ? 1 : totalCount / 10,
+    );
     const applications =
         query.data?.applicationsConnection.edges.map((edge) => edge.node) ?? [];
+    const activePage = page > totalPages ? totalPages : page;
+    const changeLimit = useCallback((val: string | null) => {
+        const parsed = val ? parseInt(val, 10) : null;
+        if (typeof parsed === "number" && !isNaN(parsed)) {
+            setLimit(parsed);
+        }
+    }, []);
 
     return (
         <>
             {isConnected && (
                 <StackApplication>
                     <Title order={4}>My Applications</Title>
+                    <Pagination
+                        styles={{ root: { alignSelf: "flex-end" } }}
+                        value={activePage}
+                        total={totalPages}
+                        onChange={setPage}
+                    />
                     <UserApplicationsTable
                         applications={applications}
                         fetching={query.fetching}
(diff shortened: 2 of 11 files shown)


apps/web/src/components/applications/applications.tsx
```diff
@@ -1,13 +1,63 @@
...
+                    py={"sm"}
+                    style={{
+                        borderTop: `calc(0.0625rem * var(--mantine-scale)) solid ${
+                            colorScheme === "light"
+                                ? "var(--mantine-color-gray-3)"
+                                : "var(--mantine-color-dark-4)"
+                        }`,
+                    }}
+                >
+                    <>
```
nevendyulgerov, line 42: No need for this extra fragment.

apps/web/src/components/applications/applications.tsx
```diff
@@ -1,13 +1,63 @@
...
+            ownerId: address?.toLowerCase(),
+        },
+    });
+    const applications =
+        query.data?.applicationsConnection.edges.map((edge) => edge.node) ?? [];
+
+    return (
+        <>
+            {isConnected && (
+                <Stack
```
nevendyulgerov, line 32: Maybe we can extract this `Stack` with its props to a separate component to prevent code duplication and use it for both `UserApplications` and `Applications`?

apps/web/src/components/entriesSummary.tsx
```diff
@@ -2,15 +2,27 @@
...
+    useStatsApplicationsOwnerQuery,
+    useStatsQuery,
+} from "../graphql/explorer/hooks/queries";
 const EntriesSummary: FC = () => {
+    const { address } = useAccount();
     const [{ data: stats }] = useStatsQuery();
+    const [{ data: applicationsOwned }] = useStatsApplicationsOwnerQuery({
+        variables: {
+            applicationId: address,
+        },
```
nevendyulgerov, line 16: Let's use the `pause` option here with `isConnected`:

```
const { address, isConnected } = useAccount();
const [query] = useStatsApplicationsOwnerQuery({
    variables: {
        ...
    },
    pause: !isConnected
});
````

We shouldn't show results for user owned applications until the user's wallet is connected.

File: apps/web/src/components/applications/applications.tsx [hunk 4]
```diff
@@ -1,13 +1,63 @@
...
+import UserApplicationsTable from "./userApplicationsTable";
+
+const UserApplications: FC = () => {
+    const { colorScheme } = useMantineColorScheme();
+    const { address, isConnected } = useAccount();
+    const [query] = useApplicationsConnectionOwnerQuery({
+        variables: {
+            orderBy: ApplicationOrderByInput.IdAsc,
+            limit: 20,
+            after: "1",
```
nevendyulgerov, line 22: Do we need the `after` variable here? This will get applications after the first one.

apps/web/src/components/entriesSummary.tsx
```diff
@@ -2,15 +2,27 @@
...
+import {
+    useStatsApplicationsOwnerQuery,
+    useStatsQuery,
+} from "../graphql/explorer/hooks/queries";
 const EntriesSummary: FC = () => {
+    const { address } = useAccount();
     const [{ data: stats }] = useStatsQuery();
+    const [{ data: applicationsOwned }] = useStatsApplicationsOwnerQuery({
+        variables: {
+            applicationId: address,
```
nevendyulgerov, line 15: Please use here:

```
applicationId: address?.toLowerCase()
```

Otherwise, you won't get correct results for the owner applications.

apps/web/src/components/applications/applications.tsx
nevendyulgerov: Another thing that I noticed but I'm not sure if it's something we should address is that on sunodo-generated dapp, the explorer api graphql (at http://localhost:8080/explorer-api/graphql) and `ApplicationsConnection` query doesn't support the `timestamp` field. So, in this case I'm getting a `400` error with message:

```
"Cannot query field \"timestamp\" on type \"Application\"."
```

Here's a screenshot:
Screenshot 2024-03-12 at 22 27 49


And here's the same but on sepolia:
Screenshot 2024-03-12 at 22 19 50

CC @brunomenezes

dandheedge: > Do we need the `after` variable here? This will get applications after the first one. Also, I see that on sepolia, certain owners (like `0xf05d57a5bed2d1b529c56001fc5810cc9afc0335`) have more than `20` applications:
> 
> Screenshot 2024-03-12 at 22 19 50
> 
> I think we'll need pagination for the `UserApplicationsTable` component.

@nevendyulgerov, initially, I want to keep it simple just to display the 20 applications a user has. Do you think we need another pagination component for UserApplicationsTable, or can we display all the data at once? My concern with adding another pagination component is that the query parameters might become bloated.

brunomenezes: > Another thing that I noticed but I'm not sure if it's something we should address is that on sunodo-generated dapp, the explorer api graphql (at http://localhost:8080/explorer-api/graphql) and ApplicationsConnection query doesn't support the timestamp field. So, in this case I'm getting a 400 error with message:

The rollups-explorer (UI and API) docker images available are probably old, but they work together. The latest ones (0.9.0 and 0.4.0) are merged on Sunodo's base code, but I am not sure if they have already released the new version, `0.11.0`. So using a local dev rollups-explorer running in parallel with sunodo-cli (i.e. sunodo run) you may have expected troubles. Here is what I do, if you are on a Mac and installed Sunodo-cli using `homebrew` you can go to `/opt/homebrew/Cellar/sunodo/SUNODO_VERSION_HERE/libexec/lib/node_modules/@sunodo/cli/dist/node/docker-compose-explorer.yaml` and replace the tags for API and UI.  In the rollups-explorer project's root you can do a `docker buildx bake` that will build a `cartesi/rollups-explorer` tag `devel` so you can add that tag in the Sunodo docker-compose file I mentioned and you have the "dev" version running if that is necessary. 


> @nevendyulgerov, initially, I want to keep it simple just to display the 20 applications a user has. Do you think we need another pagination component for UserApplicationsTable, or can we display all the data at once? My concern with adding another pagination component is that the query parameters might become bloated.

@dandheedge, Let me add two cents here. This is a list displayed "temporarily" i.e. only related to the account address of the connected wallet, so without a connected wallet it should not have a list to show. Nevertheless, pagination is important. 

that said: 
1. What would be the pros/cons of having the pagination tracked by the query-params? 
2. What would be the pros/cons of having the pagination not tracked on the query-params but only in its own local component/hook? 
3. Why would you choose one over another?

apps/web/graphql/queries.graphql
```diff
@@ -65,6 +65,12 @@ query stats($applicationId: String) {
     }
 }
 
+query statsApplicationsOwner($applicationId: String) {
+    applicationsConnection(orderBy: id_ASC, where: {owner_startsWith: $applicationId}) {
```
brunomenezes, line 69: Hey @dandheedge, The `*_startsWith` has implications for API queries. It will translate into a `Like` operator, and it makes sense when doing a query search (To some extent), but I think using `*_eq` is better here since the value is known and should `exact match`. _As a side note, as we start having more queries, we should evaluate the creation of new indexes on the API side._

apps/web/graphql/queries.graphql
```diff
@@ -105,6 +111,23 @@ query applicationsConnection($orderBy: [ApplicationOrderByInput!]!, $limit: Int,
     }
 }
 
+query applicationsConnectionOwner($orderBy: [ApplicationOrderByInput!]!, $limit: Int, $after: String, $ownerId: String) {
+    applicationsConnection(orderBy: $orderBy, first: $limit, after: $after, where: {owner_startsWith: $ownerId}) {
```
brunomenezes, line 115: Same I mentioned above regarding the `*_startsWith`

apps/web/src/components/applications/applications.tsx
dandheedge: > The rollups-explorer (UI and API) docker images available are probably old, but they work together. The latest ones (0.9.0 and 0.4.0) are merged on Sunodo's base code, but I am not sure if they have already released the new version, `0.11.0`. So using a local dev rollups-explorer running in parallel with sunodo-cli (i.e. sunodo run) you may have expected troubles. Here is what I do, if you are on a Mac and installed Sunodo-cli using `homebrew` you can go to `/opt/homebrew/Cellar/sunodo/SUNODO_VERSION_HERE/libexec/lib/node_modules/@sunodo/cli/dist/node/docker-compose-explorer.yaml` and replace the tags for API and UI. In the rollups-explorer project's root you can do a `docker buildx bake` that will build a `cartesi/rollups-explorer` tag `devel` so you can add that tag in the Sunodo docker-compose file I mentioned and you have the "dev" version running if that is necessary.

@brunomenezes here are my thoughts on your question:

#### Pagination Tracked by Query-Params

**Pros**

- Users can bookmark or share URLs that maintain the state of the application.

**Cons**

- In the `/applications` page, the page already have the query-params for **All Applications** table. Adding more pagination with tracked query-params will increase the complexity on the url and the state handling as well.

#### Pagination With Local Hook

**Pros**

- Simplifies state management within the component and results in a cleaner URL.
- [Stake Application](https://explorer.cartesi.io/) has already implemented this method, suggesting it is a reliable approach.

**Cons**

- Users cannot bookmark or share URLs that preserve the application's state.

I prefer using a local hook for pagination since the table only displays data related to the user's connected address, avoiding unnecessary complexity.

apps/web/src/components/applications/userApplicationsTable.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { FC } from "react";
+import { Application } from "../../graphql/explorer/types";
+import { ApplicationsTableProps } from "./applicationsTable";
+import UserApplicationsRow from "./userApplicationsRow";
+
+const UserApplicationsTable: FC<ApplicationsTableProps> = (props) => {
+    const { applications, fetching, totalCount } = props;
+    return (
+        <Box
+            h={"40vh"}
```
brunomenezes, line 11: @dandheedge, with that fixed, when you have no applications (my case), you have a reasonable length of empty space, pushing the `All applications`  further down without a good reason.

brunomenezes (changes requested):

brunomenezes:
@dandheedge, A few PRs were merged, could you rebase?

apps/web/src/components/applications/applications.tsx
```diff
@@ -1,11 +1,122 @@
...
+import Paginated, { perPageList } from "../paginated";
+import UserApplicationsTable from "./userApplicationsTable";
+
+type StackApplicationProps = {
+    children?: React.ReactNode;
+};
+
+/**
+ *
+ * @returns Component to wrap Application or UserApplication coponent
```
brunomenezes, line 29: small typo here

apps/web/src/components/applications/applications.tsx
```diff
@@ -1,11 +1,122 @@
...
+    const { address, isConnected } = useAccount();
+    const [page, setPage] = useState(1);
+    const [limit, setLimit] = useState<number>(10);
+    const after = page === 1 ? undefined : ((page - 1) * limit).toString();
+    const [query] = useApplicationsConnectionOwnerQuery({
+        variables: {
+            after,
+            limit,
+            orderBy: ApplicationOrderByInput.IdAsc,
+            ownerId: "0xf05d57a5bed2d1b529c56001fc5810cc9afc0335".toLowerCase(),
```
brunomenezes, line 60: @dandheedge, hardcoded address here. Currently, that is what I see when changing accounts. 👇 

my-applications-hardcoded-address

brunomenezes (changes requested):

apps/web/src/components/applications/userApplicationsRow.tsx
```diff
@@ -0,0 +1,82 @@
...
+                                    verbose: true,
+                                },
+                            )} ago`}
+                    </Text>
+                </Box>
+            </Table.Td>
+            <Table.Td>
+                <Tooltip label="Inputs">
+                    <Link
+                        href={`/applications/${appId}`}
```
brunomenezes, line 46: @dandheedge, I saw you rebase and have Neven's latest code changes related to the App's inputs. This link need to be updated accordingly.

brunomenezes (changes requested):
Rebase the branch.

Issue #127, opened 2024-03-06
Add more unit tests for transaction forms


nevendyulgerov:
## 📄 Context

Let's cover all existing transaction forms like `Ether` deposit, `ERC20` deposit, `ERC721` deposit and `Raw` input forms with as many unit tests as possible. The goal is to increase stability and safety when using those forms from the app because they make transactions and on prod it's crucial that everything works as expected for them.

## 📈 Subtasks

- [x] Add more unit tests for `Ether` deposit form
- [ ] Add more unit tests for `ERC20` deposit form
- [ ] Add more unit tests for `ERC721` deposit form
- [x] Add more unit tests for `Raw` input form
- [ ] Make sure that all forms have tests preventing issues like double hex encoding of base and exec layer data

## 🎯 Definition of Done

- [ ] All existing and new tests pass successfully.

PR #138, opened 2024-03-11
#127 Add unit tests for transaction forms

nevendyulgerov:
I added more unit tests for all transaction forms. I tried to cover all functionality and UI of those forms so we can increase stability for them.


packages/ui/src/hooks/useTokensOfOwnerByIndex.ts
dandheedge, line 1: @nevendyulgerov nice approach by making this separate hooks

dandheedge (approved):

Issue #136, opened 2024-03-09
Change cursor for color mode switch to pointer


nevendyulgerov:
## 📄 Context

When you hover over the color mode switch in the app's header the mouse cursor remains the default one. I think we should have it changed to `pointer` on hover. This will make it consistent will all other interactive elements in the header - logo, send transaction, wallet and connection sidebar.

![color-mode-switch](https://github.com/cartesi/rollups-explorer/assets/6005179/7757a807-f62e-4153-90ed-fb3d41b44533)

## ✔️ Solution

We can use [this guide](https://mantine.dev/core/switch/#pointer-cursor) to enable the `pointer` cursor.

## 🎯 Definition of Done

- [ ] On hover the color mode switch cursor becomes `pointer`.

Issue #134, opened 2024-03-09
Search query is not propagated on page refresh


nevendyulgerov:
## 🙂 Expected behavior

Search queries made through the `Search by Address / Txn Hash / Index` search should be propagated on page refresh. This allows users to share links from the app, that when pasted in the browser's url bar will open the same page with same results for both users.

## 🫠 Actual behavior

When using the `Search by Address / Txn Hash / Index` search, the inserted query is not propagated on page refresh. As a result the search input is not pre-filled with the query and no search is being made.

## 🧪 Minimal test case

1. Go to `/inputs` page
2. Search for some inputs
3. Notice that the url is changed and `query={search}` is being appended to it.
4. Refresh the page
5. Notice the the url changes and the `query={search}` is removed. Also no search is being made and no relevant results are displayed.

## 🌎 Environment

Any network and any rollups version.

## Video demonstrating the issue:

https://github.com/cartesi/rollups-explorer/assets/6005179/1d90afe1-da9c-4ff2-abf5-6ed37c0b722f

PR #139, opened 2024-03-11
Fix Page Input No Result Display and Query Missing on Refresh

### LINKED ISSUE #133: No results label is not centered for inputs table
## 🙂 Expected behavior

The no results label for inputs table should be centered.

## 🫠 Actual behavior

The no results label for inputs table is not centered. Here's a screenshot:
no-results-label

## 🧪 Minimal test case

1. Go to `/inputs` page
2. Insert a search query that results in no entries being found
3. Observe the empty table and the no results label in it

## 🌎 Environment

Any network and any rollups version.

dandheedge:
Close #133 #134 #136

### DIFF
--- a/apps/web/src/components/search.tsx
+++ b/apps/web/src/components/search.tsx
@@ -9,7 +9,7 @@ export type SearchProps = {
 
 const Search: React.FC<SearchProps> = ({ onChange, isLoading }) => {
     const { query, updateQueryParams } = useQueryParams();
-    const [keyword, setKeyword] = useState<string>("");
+    const [keyword, setKeyword] = useState<string>(query);
 
     useEffect(() => {
         updateQueryParams(keyword);
--- a/apps/web/src/components/applications/applicationsTable.tsx
+++ b/apps/web/src/components/applications/applicationsTable.tsx
@@ -2,9 +2,9 @@
 
 import { Loader, Table } from "@mantine/core";
 import { FC } from "react";
-import ApplicationRow from "./applicationRow";
 import { ApplicationItemFragment } from "../../graphql/explorer/operations";
 import { Application } from "../../graphql/explorer/types";
+import ApplicationRow from "./applicationRow";
 
 export interface ApplicationsTableProps {
     applications: ApplicationItemFragment[];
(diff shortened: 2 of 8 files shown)


apps/web/src/components/search.tsx
```diff
@@ -9,7 +9,7 @@ export type SearchProps = {
 
 const Search: React.FC<SearchProps> = ({ onChange, isLoading }) => {
     const { query, updateQueryParams } = useQueryParams();
-    const [keyword, setKeyword] = useState<string>("");
+    const [keyword, setKeyword] = useState<string>(query);
```
nevendyulgerov, line 12: This persists the query and fetches the correct data based on the query, however sometimes it can be noticed that actually two requests are being made for the UI to show the filtered data:

https://github.com/cartesi/rollups-explorer/assets/6005179/2f66bd51-3aa0-4181-be20-34f51e502888

The first is with an empty query and the second with the one from the url. We can mitigate this by slightly extending the `Inputs` component:

```
const Inputs: FC<InputsProps> = ({
    orderBy = InputOrderByInput.TimestampDesc,
    applicationId,
}) => {
    const { query: urlQuery } = useQueryParams();
    const [query, setQuery] = useState(urlQuery);
    ...
```
This way, the initial request for inputs will always have the url query. The resulting UX is this:

https://github.com/cartesi/rollups-explorer/assets/6005179/2fb5bf5b-d6f4-42e6-bc99-7a15d90d5bcd

This way, the second request will hit the graphql cache (because the first and second query will be the same) and data will be pulled from it instead. The win is slightly better UI and one less request for the end user.

nevendyulgerov: Let's also please add unit tests for this functionality to be on the safe side.

nevendyulgerov (commented):
Nice work, @dandheedge ! We can do a couple of minor tweaks here and there to make some small improvements to the UI.

nevendyulgerov (approved):
Good stuff, @dandheedge, thanks for making those tweaks!

brunomenezes (approved):
LGTM. I did a quick rebase before merging.

PR #143, opened 2024-04-03
Feature: ERC-1155 Transfers

brunomenezes:
### Summary
code changes to support multi-token transfers through available portals [ERC-1155 Batch Portal](https://next.docs.cartesi.io/cartesi-rollups/api/json-rpc/portals/ERC1155BatchPortal/) and [ERC-1155 Single Portal](https://next.docs.cartesi.io/cartesi-rollups/api/json-rpc/portals/ERC1155SinglePortal/). 

I will get ahead and explain some rationale for the decisions I made. Note that only one component is exported, but based on the mode, it will pick one form or the other. There are shared components and validators where due, but the amount of differences in what makes one or the other form ready and given the way the `hooks` work would increase the difficulty in **maintenance** (The most critical piece to me. Developing is easy, maintaining is hard) given the number of variables to look after to enable or not these calls (only because of this eager nature of hooks), also other logical variables to decide when action buttons should be enabled or not that would be another unnecessary complexity. _So there you have it, dear reviewer(s)_

The implementation was done keeping in mind the specification [EIP-1155](https://eips.ethereum.org/EIPS/eip-1155). At the moment, the only thing ruled out is the possible use of `locale` in the `URI` for `i18n/i10n`.

### Summary of changes: 
* Added check for supported interfaces in a given contract address before allowing the user to proceed (Check the spec) 
* Added validations to check the operator's (Portal) approval and to see if the user is not exceeding the balance. In batch mode, any deposit in the review list is also included in a sum to be checked.
* I added a metadata preview when available. If not, informational messages are displayed, but the user can proceed if comfortable.
* Added a new graphql query to be able to search for already indexed ERC-1155 contracts.
* Added a new component to generate a link to the connected chain block-explorer to check the transaction upon transfer success (confirmation). _Note: devnet does not have one, so nothing is generated_
* Added test cases.


apps/web/src/components/BlockExplorerLink.tsx
```diff
@@ -0,0 +1,45 @@
...
+ * Works in conjuction with Wagmi. It requires a Wagmi-Provider to work as expected.
+ * When running devnet it will not render a block-explorer link.
+ *
+ */
+export const BlockExplorerLink: FC<BlockExplorerLinkProps> = ({
+    value,
+    type,
+}) => {
+    const config = useConfig();
+    const explorerUrl = config.chains[0].blockExplorers?.default.url;
```
fabiooshiro, line 25: it always have at least one item?

brunomenezes: Yes! It is the chain set for the running project [here](https://github.com/cartesi/rollups-explorer/blob/main/apps/web/src/providers/walletProvider.tsx#L25C7-L25C12). As CartesiScan only works with one of the following [`mainnet`, `sepolia`, `local`]. So, that `config.chains` will have one item at all times.

brunomenezes:
nudging reviewers: @nevendyulgerov , @dandheedge , @tuler

apps/web/test/components/BlockExplorerLink.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+            chains: [sepolia],
+        });
+    });
+
+    afterEach(() => {
+        vi.clearAllMocks();
+        cleanup();
+    });
+
+    it("it should render nothing when the block-explorer URL is not available", () => {
```
nevendyulgerov, line 28: Maybe we can omit the `it` from the test name as done for all other tests?

eg `it('should....')

apps/web/src/components/sendTransaction.tsx
```diff
@@ -48,6 +55,33 @@ const SendTransaction: FC<DepositProps> = ({
...
+                withCloseButton: true,
+                autoClose: false,
+                color: "green",
+                ...message,
+            });
+
+            setMultiTokenId("");
+            setApplicationId("");
+        },
+        [],
```
nevendyulgerov, line 82: Missing dependency here - `setMultiTokenId`.

packages/ui/src/ERC1155DepositForm/DepositFormBatch.tsx
```diff
@@ -0,0 +1,415 @@
...
+        address: erc1155Contract.address,
+        args: [getAddress(address!), tokenId!],
+        query: {
+            enabled: tokenId !== undefined,
+        },
+    });
+
+    const supportsInterface = useReadErc1155SupportsInterface({
+        address: erc1155Contract.address,
+        args: ["0xd9b67a26"],
```
nevendyulgerov, line 161: Maybe we should add a comment here explaining what this arg means? Something like `ERC-165 identifier`.

packages/ui/src/ERC1155DepositForm/DepositFormSingle.tsx
```diff
@@ -0,0 +1,393 @@
...
+        address: erc1155Contract.address,
+        args: [getAddress(address!), tokenId!],
+        query: {
+            enabled: tokenId !== undefined,
+        },
+    });
+
+    const supportsInterface = useReadErc1155SupportsInterface({
+        address: erc1155Contract.address,
+        args: ["0xd9b67a26"],
```
nevendyulgerov, line 138: Maybe we should add a comment here explaining what this arg means? Something like `ERC-165 identifier`.

nevendyulgerov (approved):
LGTM!

apps/web/src/components/sendTransaction.tsx
brunomenezes: I look inside the implementation for the setter coming from the useDebouncedState and it does not guarantee the identity of the function. So, I replaced it with the useState and just used the **useDebouncedValue** for derivation. Good catch.

brunomenezes: [4aa255e](https://github.com/cartesi/rollups-explorer/pull/143/commits/4aa255ed5278ecdf7a30524696df46661572248f)

packages/ui/src/ERC1155DepositForm/DepositFormSingle.tsx
brunomenezes: [0ad2fee](https://github.com/cartesi/rollups-explorer/pull/143/commits/0ad2fee46a44a3e1e73638ed2ca9eea82b53c089)

packages/ui/src/ERC1155DepositForm/DepositFormBatch.tsx
brunomenezes: [0ad2fee](https://github.com/cartesi/rollups-explorer/pull/143/commits/0ad2fee46a44a3e1e73638ed2ca9eea82b53c089)

brunomenezes: [c6d0f34](https://github.com/cartesi/rollups-explorer/pull/143/commits/c6d0f34b51a422b8872718cd51adabf505e28282)

nevendyulgerov (approved):
LGTM, thanks for making those tweaks!

Issue #147, opened 2024-04-04
Upgrade nextjs package to latest version


nevendyulgerov:
## 📄 Context

NextJS is now one minor version ahead - https://www.npmjs.com/package/next. Let's upgrade to it.

Here's a [post](https://nextjs.org/blog/next-14-1) about v14.1.

## 📈 Subtasks
- [x] Make sure dependencies are updated in the lock file (package-lock.json).
- [x] Verify whether everything is working as expected.
- [x] Test the app via docker to ensure it works as expected.

PR #148, opened 2024-04-07
#147 Upgrade nextjs to latest version

nevendyulgerov:
I upgraded next to the latest version and tested the app afterwards. I didn't notice any issues and according to the [docs](https://nextjs.org/blog/next-14-1) most changes affect self-hosting, dev builds performance, dx, parallel/intercepted routes so I think we are safe to upgrade.

Looking at the next repo I did notice a couple of issues reported related to 14.1 - https://github.com/vercel/next.js/issues/61002 and https://github.com/vercel/next.js/issues/62727. However, both seem to be resolved in 14.1.4.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -28,7 +28,7 @@
         "graphql": "^16",
         "graphql-tag": "^2",
         "lokijs": "^1",
-        "next": "^14.0.3",
+        "next": "^14.1.4",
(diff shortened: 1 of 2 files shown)


nevendyulgerov:
Hey @brunomenezes , I tested this one via the docker container and didn't notice any issues. The app works as expected.

Issue #150, opened 2024-04-09
Add connections page


brunomenezes:
## 📄 Context
Today, the connections list lives inside an `<aside>` visible when clicking the vertical dots icon in the header. The intention is to allow the connection to have its page.

## ✔️ Solution
Move the connection to its page and remove the `<aside />` and all mechanisms necessary for it to work.

* Remove the vertical dots (a.k.a vertical kebab icon) and also the whole aside area. 
* Create a collapsable Settings with a connection link to the new page (**Settings Icon [TbAdjustmentsHorizontal](https://react-icons.github.io/react-icons/search/#q=TbAdjustmentsHorizontal)**). 
* Settings is a virtual grouping, i.e. no /settings route (For now). 
* Implement a connections page (/connections).
* List the connection cards.

## 📈 Subtasks

- [x] Remove `<aside />` and the mechanism to make it work.
- [x] Add Settings group with connection as sub-item.
- [x] Add new connection page under `/connections` route.
- [x] Remove collapse and scroll area from `ConnectionView` component
- [x] Remove obsolete tests for `ConnectionView` component
- [x] Add tests for Settings group

## 🎯 Definition of Done

- [x] The previous view is removed, including all triggers to it. i.e. `<aside>`
- [x] Can list all the created connections under `/connections`.
- [x] Can create a connection from `/connections`.
- [x] All tests are passing.

PR #157, opened 2024-04-11
#150 Add connections page

nevendyulgerov:
I added a settings group to the navbar. It contains a single nav link to the connections page. I also removed a couple of obsolete tests for `ConnectionView` component and added several for the `Shell` component.

### DIFF
--- a/apps/web/src/components/connection/connectionInfo.tsx
+++ b/apps/web/src/components/connection/connectionInfo.tsx
@@ -21,7 +21,7 @@ const ConnectionInfo: FC<ConnectionInfoProps> = ({ connection }) => {
     const { removeConnection } = useConnectionConfig();
     const theme = useMantineTheme();
     return (
-        <Card withBorder py="lg" my="md" radius="sm" shadow="sm">
+        <Card withBorder py="lg" mt="md" radius="sm" shadow="sm">
             <Card.Section inheritPadding withBorder>
                 <Flex justify="space-between">
                     <Address value={connection.address} shorten icon />
--- a/apps/web/src/components/connection/connectionInfo.tsx
+++ b/apps/web/src/components/connection/connectionInfo.tsx
@@ -21,7 +21,7 @@ const ConnectionInfo: FC<ConnectionInfoProps> = ({ connection }) => {
     const { removeConnection } = useConnectionConfig();
     const theme = useMantineTheme();
     return (
-        <Card withBorder py="lg" mt="md" radius="sm" shadow="sm">
+        <Card withBorder py="lg" radius="sm" shadow="sm">
             <Card.Section inheritPadding withBorder>
                 <Flex justify="space-between">
                     <Address value={connection.address} shorten icon />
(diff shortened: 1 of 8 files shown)


apps/web/src/components/connection/connectionView.tsx
```diff
@@ -1,24 +1,20 @@
...
-import { useDisclosure, useViewportSize } from "@mantine/hooks";
-import { TbChevronDown, TbChevronUp, TbPlus } from "react-icons/tb";
+import { TbPlus } from "react-icons/tb";
 import { useConnectionConfig } from "../../providers/connectionConfig/hooks";
 import ConnectionInfo from "./connectionInfo";
 
 const ConnectionView = () => {
-    const [showConnectionList, { toggle: toggleConnectionList }] =
-        useDisclosure(true);
     const { showConnectionModal } = useConnectionConfig();
```
dandheedge, line 17: I think the `showConnectionModal` and `listConnections` could be on the same line

```suggestion
    const { showConnectionModal, listConnections } = useConnectionConfig();
```

apps/web/src/components/layout/shell.tsx
dandheedge, line 1: https://github.com/cartesi/rollups-explorer/blob/75e74e818a72c9a3155779618d124ccc42b22f55/apps/web/src/components/layout/shell.tsx#L56

We can enable the transition by using this

```
const { colorScheme, toggleColorScheme } = useMantineColorScheme({
        keepTransitions: true,
    });
```

nevendyulgerov: Nice, it's much better with transitions after this change :)

apps/web/src/components/connection/connectionInfo.tsx
```diff
@@ -21,7 +21,7 @@ const ConnectionInfo: FC<ConnectionInfoProps> = ({ connection }) => {
     const { removeConnection } = useConnectionConfig();
     const theme = useMantineTheme();
     return (
-        <Card withBorder py="lg" my="md" radius="sm" shadow="sm">
+        <Card withBorder py="lg" mt="md" radius="sm" shadow="sm">
```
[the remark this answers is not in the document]
nevendyulgerov: I assume we are good with this `mt` setting or you suggest that I stick to `my`?

brunomenezes (commented):
Hey @nevendyulgerov, I had a run with the page. I think the `ConnectionView` could be updated from a `Stack` approach to using a `Grid` to better utilize the new space; before the space was smaller, so it was the only sensible choice. Thinking of breaking points, it could have 1 column as a **base**, and anything from **md** and above goes with two columns. What do you think?

apps/web/src/components/connection/connectionInfo.tsx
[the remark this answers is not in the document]
dandheedge: After reviewing the UI again, I find it more balanced to use mt on this component. Sorry for the confusion, @nevendyulgerov.

brunomenezes (approved):
@nevendyulgerov, I rebased and pushed a small change to the menu for the small screen; it was not closing the sub-menu when clicking.

Issue #152, opened 2024-04-09
Add new route for application's input


brunomenezes:
## 📄 Context
We will add more ways to access different types of output from the application's backend. Therefore, the content in the application's `Home` should be moved to a matching route i.e. `/applications/<address>/inputs`. That will give space and granularity to explore other views on specific outputs when available e.g. notices/vouchers/reports. 

_PS: That will serve as a reference for the output unification that will come with rollups V2._

## ✔️ Solution
Move the content in the `/applications/<address>` to the matching route.

* Move the current application's `Home`  content to `/applications/<address>/inputs`
* Update the input's table link for the `application` address to the new URL.
* Update the `Home` page cards application link to point to the new URL.
* Update the `Inputs` action button in the applications table to the new URL i.e. located in `/applications`

## 📈 Subtasks

- [x] Move application inputs content to the new URL
- [x] Update the Home page cards links to the new URL format for Application addresses.
- [x] Update application links format in the `Inputs` page. 
- [x] Update application links format in the `Home` page.
- [x] Update tests (if required)
- [x] Update stories (if required)

## 🎯 Definition of Done

- [x] The application input list is under `/applications/<address>/inputs`
- [x] Clicking the application address in the inputs table goes to the specific application input list. 
- [x] Clicking the application address in any of the `Home` page summary cards goes to the application's input list.
- [x] Clicking the the inputs action button in the list of applications goes to the correct application input list page.
- [ ] CI is passing all the checks
- [ ] Vercel build is passing.

PR #159, opened 2024-04-12
#152 Add route for applications inputs

nevendyulgerov:
I added the new `/applications/{address}/inputs` route and updated links that previously pointed to `/applications/{address}` to now point to that new route.


brunomenezes (changes requested):
Hi @nevendyulgerov, it looks good. You just forgot to remove the page [here](https://github.com/cartesi/rollups-explorer/blob/3ce090cbd5a3a39addf3bb4c979ac78be8cab544/apps/web/src/app/applications/%5Baddress%5D/page.tsx), currently I can see the same content if I remove the `/inputs` route at the end, should be the usual `404` not-found message. 👇

brunomenezes (approved):

PR #161, opened 2024-04-22
#156 Voucher content decoding

### LINKED ISSUE #156: Voucher content decoding
## 📄 Context
Given that a voucher has a destination (Address) and the payload (Encoded arguments), we can recover the ABI of this destination address and use it to try to parse the encoded payload (i.e. the input for the destination) for this type of output.

**Ref Rollups Contracts:** [CartesiDApp#executeVoucher](https://github.com/cartesi/rollups-contracts/blob/release/1.2.x/onchain/rollups/contracts/dapp/CartesiDApp.sol#L137C25-L137C37)

## ✔️ Solution
We can leverage two tools here. We could experiment with [WhatsAbi](https://github.com/shazow/whatsabi) to retrieve the ABI from an Address (Check the documentation for more details on usage), and then use [Viem](https://viem.sh) to decode the `payload` and display it in the Input-details box. 

* Add WhatsAbi as a new Dependency
* Add logic to retrieve the ABI from a contract. Handle any possible failures gracefully, The Application should continue to work in case it fails to retrieve information. 
* If the ABI information is valid, try to decode using Viem. Handle any possible errors during the decoding process. 
* If the ABI retrieval or decoding fails, send down to the `input details` the usual **encoded payload**.
* Failures in any of the steps above should be silent i.e. No notifications/feedback. That is an enhancement of visibility, and if it is not available, we just move on (no fuss about it).
* For decoding success, there are two options (pick one)
  * Get the decoded content, transform it into a Hex value, pass it down and let the `Content.tsx` do what it is doing.
  * The other option is to adjust the code/types all the way to the `Content.tsx` to recognise what it is receiving and apply the correct transformation functions. For that option, everything else should still work as expected. 
  
 
PS: The first option is more sensible in the decoding success section above. Given the number of unrelated changes and checks required by the second option. It can be a separate `refactor` issue if necessary.

## 📈 Subtasks

- [x] Implement logic to retrieve the ABI from a destination address.
- [x] Implement logic to decode the payload.
- [x] Make the adjustments to display the payload.
- [x] Add test cases. 

## 🎯 Definition of Done

- [x] When voucher information is valid, I can see the payload in a human-readable form when using the current decodes (`Text`, `JSON`)
- [x] When voucher information is invalid, it continues to work as it did previously.
- [x] CI checks are passing.

dandheedge:
Close #156


brunomenezes:
@dandheedge how is that going? 

@nevendyulgerov is the **drawingcanvas** you productiv-e deployed generating vouchers? is that in Sepolia?

packages/ui/src/hooks/useDecodeVoucherPayload.ts
```diff
@@ -0,0 +1,40 @@
...
+            } finally {
+                setLoading(false);
+            }
+        };
+        if (typeof content !== "string") {
+            decodePayload();
+        } else {
+            setData(content);
+        }
+    }, [destination, payload]);
```
nevendyulgerov, line 37: Missing dependency `content` here.

packages/ui/src/hooks/useDecodeVoucherPayload.ts
```diff
@@ -0,0 +1,40 @@
...
+import { useEffect, useMemo, useState } from "react";
+export const useDecodeVoucherPayload = (
+    content: string | DecodeVoucherPayloadParamsType,
+) => {
+    const { destination, payload } =
+        content as unknown as DecodeVoucherPayloadParamsType;
+    const [data, setData] = useState<unknown | string>("");
+    const [loading, setLoading] = useState<boolean>(true);
+    const [error, setError] = useState<Error | null>(null);
+    useEffect(() => {
```
nevendyulgerov, line 14: I would refactor this a bit to decouple the effect and the `decodePayload` function:

```
const decodePayload = useCallback(async () => {
    setLoading(true);
    try {
        const decodedData = await decodeVoucherPayload({
            destination,
            payload,
        });
        setData(decodedData);
    } catch (err) {
        setError(
            err instanceof Error
                ? err
                : new Error(
                      "An error occurred while decoding the voucher payload",
                  ),
        );
        setData(payload); // Fallback value for raw value.
    } finally {
        setLoading(false);
    }
}, [destination, payload]);

useEffect(() => {
    if (typeof content === "string") {
        setData(content);
    } else {
        decodePayload().catch(() => undefined);
    }
}, [content, decodePayload, destination, payload]);
```

Please note the error message (I tweaked it a bit to make it more detailed) and also I added a `catch` to the `decodePayload()` invocation (I was getting a warning about ignoring the returned promise. 

Finally, in the effect, I'm checking if the content is of type string, not the other way around. I find it easier to write/read conditions starting with the positive case.

packages/ui/src/InputDetails/VoucherContent.tsx
```diff
@@ -11,7 +11,6 @@ const DISPLAY_NAME = "VoucherContent" as const;
 const VoucherContent: FC<VoucherContentType> = (props) => {
     const { children = null, ...restProps } = props;
     PageableContent.displayName = DISPLAY_NAME;
-
```
nevendyulgerov, line 14: I don't think we need to touch this file if no code is changed in it. Please revert this change to reduce the PR's footprint.

packages/ui/src/InputDetails/Content.tsx
```diff
@@ -37,7 +55,61 @@ export const ContentTypeControl = ({
     );
 };
 
-export const DisplayContent = cond<
+const LoadingState: React.FC = () => {
```
nevendyulgerov, line 58: I'm getting a `Referenced UMD global variable` warning here, please import `FC` from `react` package instead and then refactor like this:

```
const LoadingState: FC = () => {...}
```

packages/ui/src/InputDetails/Content.tsx
```diff
@@ -37,7 +55,61 @@ export const ContentTypeControl = ({
...
+    return (
+        <Stack>
+            <Skeleton animate={false} height={8} radius="xl" />
+            <Skeleton animate={false} height={8} mt={6} radius="xl" />
+        </Stack>
+    );
+};
+
+export const VoucherDisplayContent: React.FC<VoucherDisplayContentProps> = ({
```
nevendyulgerov, line 67: Same for `React.FC` as above.

packages/ui/src/InputDetails/Content.tsx
```diff
@@ -37,7 +55,61 @@ export const ContentTypeControl = ({
...
+    type,
+    content,
+}: {
+    type: ContentType;
+    content: DecodeVoucherPayloadParamsType;
+}) => {
+    const { data, loading, error } = useDecodeVoucherPayload(content);
+    if (loading) return <LoadingState />;
+    if (type === "json") {
```
nevendyulgerov, line 76: Instead of having multiple return statements I'd simplify this to:

```
return (
    <>
        {loading ? (
            <LoadingState />
        ) : type === "json" ? (
            <JsonInput
                rows={10}
                value={
                    (error as unknown as string) ??
                    (JSON.stringify(data) as string)
                }
                readOnly
                placeholder="No content defined"
            />
        ) : type === "text" ? (
            <Textarea
                rows={10}
                value={
                    (error as unknown as string) ??
                    (JSON.stringify(data) as string)
                }
                readOnly
                placeholder="No content defined"
            />
        ) : (
            <Textarea
                rows={10}
                value={content.payload as Hex}
                readOnly
                placeholder="No content defined"
            />
        )}
    </>
);
```

Single return statement and the conditions are baked directly in the jsx code.

packages/ui/src/InputDetails/Content.tsx
```diff
@@ -75,3 +147,11 @@ export const DisplayContent = cond<
         ),
     ],
 ]);
+
+export const DisplayContent: React.FC<FCContentType> = ({ type, content }) => {
```
nevendyulgerov, line 151: Same for `React.FC` as above. Also, here we can refactor the logic slightly:

```
return (
    <>
        {typeof content === "string" ? (
            <DefaultDisplayContent type={type} content={content} />
        ) : (
            <VoucherDisplayContent type={type} content={content} />
        )}
    </>
);
```

packages/ui/src/InputDetails/Content.tsx
```diff
@@ -37,7 +55,61 @@ export const ContentTypeControl = ({
...
+    content: DecodeVoucherPayloadParamsType;
+}) => {
+    const { data, loading, error } = useDecodeVoucherPayload(content);
+    if (loading) return <LoadingState />;
+    if (type === "json") {
+        return (
+            <JsonInput
+                rows={10}
+                value={
+                    (error as unknown as string) ??
```
nevendyulgerov, line 81: Can we just do this instead:

```
value={error?.toString() ?? JSON.stringify(data)}
```

packages/ui/src/InputDetails/Content.tsx
```diff
@@ -37,7 +55,61 @@ export const ContentTypeControl = ({
...
+                placeholder="No content defined"
+            />
+        );
+    }
+    if (type === "text") {
+        return (
+            <Textarea
+                rows={10}
+                value={
+                    (error as unknown as string) ??
```
nevendyulgerov, line 94: Same as above for the `value` prop.

packages/decoder/src/index.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Asynchronously decodes a voucher payload using the provided ABI.
+ *
+ * @param destination - The destination address to be executed, expected to be a string prefixed with '0x'.
+ *  @param params - The payload to decode, expected to be a string prefixed with '0x'.
+ * @returns The decoded payload or null if an error occurs.
+ */
+
+export type DecodeVoucherPayloadParamsType = {
+    destination: `0x${string}`;
```
nevendyulgerov, line 14: Please use `Address` from `viem` instead for those types.

packages/decoder/src/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    functionName: string;
+    args: unknown[] | undefined;
+};
+
+export const decodeVoucherPayload = async ({
+    destination,
+    payload,
+}: DecodeVoucherPayloadParamsType): Promise<DecodeVoucherPayloadReturnType> => {
+    const provider = await getProvider();
+    try {
```
nevendyulgerov, line 28: No need for this `try...catch` block here. Let's leave the error handling to the callee (`useDecodeVoucherPayload`) instead.

packages/decoder/src/index.ts
```diff
@@ -0,0 +1,49 @@
...
+        // Fetch the ABI using the destination address
+        const destinationAddressAbi = await whatsabi.autoload(
+            destination as string,
+            { provider },
+        );
+
+        // Decode the payload using the fetched ABI
+        const res = await decodeFunctionData({
```
nevendyulgerov, line 36: `decodeFunctionData` is not `async`, thus we don't need the `await` keyword here.

packages/decoder/src/index.ts
```diff
@@ -0,0 +1,49 @@
...
+            abi: destinationAddressAbi.abi,
+            data: payload as `0x${string}`,
```
nevendyulgerov, line 38: No need for type assertion here.

packages/decoder/test/decodeVoucherPayload.test.ts
```diff
@@ -0,0 +1,65 @@
+import { autoload, whatsabi } from "@shazow/whatsabi";
+import { ethers } from "ethers";
+import { decodeFunctionData } from "viem";
+import { describe, expect, it, vi } from "vitest";
+import { decodeVoucherPayload } from "../src/index";
```
nevendyulgerov, line 5: Import can be shortened to `../src`

nevendyulgerov (changes requested):
I left some comments with suggestions for general improvements. I'll do another round of testing of the actual feature later today and tomorrow.

brunomenezes:
I will take over that one.

Issue #166, opened 2024-04-30
Add test cases for the Navigation menu.


nevendyulgerov:
## 📄 Context

Add e2e tests for the navigation menu.

## 📈 Subtasks

- [x] Add test for home page navigation
- [x] Add test for applications page navigation
- [x] Add test for inputs page navigation
- [x] Add test for settings - connection page navigation

## 🎯 Definition of Done

- [x] All new and existing e2e tests are passing

PR #167, opened 2024-04-30
#166 Add e2e tests for navigation

nevendyulgerov:
I added e2e tests for the navigation.


apps/web/e2e/menu.spec.ts
```diff
@@ -1,89 +1,194 @@
...
-        await expect(
-            page.getByPlaceholder("Search by Address / Txn Hash / Index"),
-        ).toBeVisible();
+        // Verify page heading
+        await expect(getPageTitle(page, "Inputs")).toBeVisible();
 
+        // Verify that inputs search is visible
         await expect(
-            page.getByRole("heading", { name: "Inputs" }),
+            page.getByPlaceholder("Search by Address / Txn Hash / Index"),
```
dandheedge, line 58: @nevendyulgerov I think we can use `getByTestId` as a `FrameLocator` to achieve more precise element selection and bulletproof testing. Searching by a long placeholder, such as Search by Address / Txn Hash / Index, is more prone to breaking due to text changes, such as accidental whitespace or typos.

dandheedge (changes requested):
Here also

https://github.com/cartesi/rollups-explorer/blob/a23c34e5f2b0581b5b2afc148d31e19dc7d32b4c/apps/web/e2e/menu.spec.ts#L153

dandheedge:
@nevendyulgerov LGTM, only need some rebase on `apps/web/src/app/applications/page.tsx` and we can merge it

brunomenezes (approved):

apps/web/e2e/menu.spec.ts
brunomenezes: Just for thought: it is not bad to get by the placeholder as it is something that the user can "see".

Issue #168, opened 2024-04-30
Add test cases for the Home page.


nevendyulgerov:
## 📄 Context

Add e2e tests for the home page.

## 📈 Subtasks

- [x] Add test for the inputs summary card
- [x] Add test for the applications summary card
- [x] Add test for the latest inputs table
- [x] Add test for the latest applications table

## 🎯 Definition of Done

- [x] All new and existing e2e tests are passing

PR #169, opened 2024-04-30
#168 Add e2e tests for home page

nevendyulgerov:
I added e2e tests for the home page covering the summary cards for inputs and applications as well as the latest inputs and applications tables.

### DIFF
--- a/apps/web/src/components/entriesSummary.tsx
+++ b/apps/web/src/components/entriesSummary.tsx
@@ -21,11 +21,13 @@ const EntriesSummary: FC = () => {
         : 0;
 
     return (
-        <Summary
-            inputs={stats?.inputsConnection?.totalCount ?? 0}
-            applications={stats?.applicationsConnection?.totalCount ?? 0}
-            applicationsOwned={applicationsOwnedCount}
-        />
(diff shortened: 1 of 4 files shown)


apps/web/src/components/entriesSummary.tsx
```diff
@@ -21,11 +21,13 @@ const EntriesSummary: FC = () => {
         : 0;
 
     return (
-        <Summary
-            inputs={stats?.inputsConnection?.totalCount ?? 0}
-            applications={stats?.applicationsConnection?.totalCount ?? 0}
-            applicationsOwned={applicationsOwnedCount}
-        />
+        <div data-testid="entries-summary">
```
brunomenezes, line 24: Hi @nevendyulgerov,  Does it not work adding the attribute `data-testid` in the `<Summary />` rather than adding another DOM node just for the E2E?  I assume you use that to scope the tree during the element search. If that does not work out of the box, maybe consider extending the `SummaryProps` API to accept these `data-*` properties since this component is "generic". What do you think?

dandheedge (approved):
LGTM!

Issue #172, opened 2024-04-30
Add content filtering by tabs in the applications page


brunomenezes:
## 📄 Context
The application's page (i.e. `/applications`) now lists all applications and the applications belonging to the connected account. Because both have title+pagination-buttons+table, the page became noisy with too much. To return to more driven content, tabs can be added to separate one from the other. 

## ✔️ Solution
Implement two tabs to separate the content listing all the applications from the owned ones. 

* Add two tabs located below the main page title. 
* The tabs should be theme controlled. i.e. Using the default appearance so theme changes affect everywhere accordingly.
* The copy for the tabs .e.g `All Apps` and `My Apps`
* Remove the sub-titles as they become a redundancy, i.e. `My Applications` and `All Applications`
* The default selected tab is the `All Apps`.

## 📈 Subtasks

- [x] Implement the tabs logic. 

## 🎯 Definition of Done

- [x] The user can see the list of all the apps when accessing the applications page. 
- [x] The user can select the tab to see the apps the connected account owns.
- [x] Add unit tests.

PR #174, opened 2024-05-05
#172 Add tabs to applications page

nevendyulgerov:
I added two tabs to the applications page to respectively display all apps table and my apps table. I also refactored slightly the `UsersApplications` component to utilise the `Paginated` component. Additionally, I covered the new logic with unit tests.

### DIFF
--- a/apps/web/src/components/applications/applications.tsx
+++ b/apps/web/src/components/applications/applications.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { Tabs } from "@mantine/core";
-import { FC, useCallback, useEffect, useState } from "react";
+import { FC, useCallback, useState } from "react";
 import { useAccount } from "wagmi";
 import {
     useApplicationsConnectionOwnerQuery,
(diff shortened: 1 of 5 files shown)


apps/web/src/components/applications/applications.tsx
```diff
@@ -134,23 +65,53 @@ const Applications: FC = () => {
...
+            {isConnected && (
+                <Tabs value={activeTab} onChange={onChangeTab}>
+                    <Tabs.List>
+                        <Tabs.Tab value="all-apps">All Apps</Tabs.Tab>
+                        <Tabs.Tab value="my-apps">My Apps</Tabs.Tab>
+                    </Tabs.List>
+                </Tabs>
+            )}
+
+            {activeTab === "all-apps" ? (
```
brunomenezes, line 110: Hi @nevendyulgerov, you could use the Tabs Panel component to do the switch rather than doing it manually. Have a look [here](https://mantine.dev/core/tabs/#usage)

brunomenezes: Looking closer, I notice that if not connected, you're hiding the tabs, but that isn't necessary. 

I see a couple of alternatives: 

1. Display a different message if not connected in the table row, e.g. "Connect your wallet to list your Applications."
2. Disable the Tab and display a tooltip with a message regarding the need to connect the wallet. 

The **second** is not bad, but I vote for the **first** since it is permissive and explains what needs to be done to load some particular data.

brunomenezes (approved):

PR #176, opened 2024-05-06
Feat: Application summary page access

brunomenezes:
### Summary
Code changes to add access to an application's summary page from the `/applications` page. Also, include a small change to avoid unintended rewrites in the query parameters done by the component not selected, i.e., the only mounted component is the one under the selected tab.


brunomenezes:
I will merge that because it locks the `All Apps` tab on page one. I did not notice that during the test/review. 

cc: @nevendyulgerov , @dandheedge

Issue #179, opened 2024-05-09
Implement IndexedDB for storing connections


nevendyulgerov:
## 📄 Context

Connections, created by the user, are currently stored in `localStorage`. Data from this storage can be easily cleared thus removing any saved connections, which may be confusing for the user. 

On the other hand, IndexedDB provides a more robust and strict way for removing data. So, implement IndexedDB as a primary data storage layer for the connections. In case IndexedDB is not supported in the user's browser, utilise `localStorage` as a fallback.

## ✔️ Solution

Connections are stored in IndexedDB. If that is not supported,`localStorage` is used.

## 🎯 Definition of Done

- [ ] Connections are displayed as before but stored in IndexedDB when that's available.
- [ ] When IndexedDB is not available, the connections are stored in `localStorage`.

brunomenezes:
I am not seeing here, so one thing to notice and include as a requirement for good interoperability is that in case the `indexedDB` is supported; do a quick check in the `localStorage` and preload any existing connection, so they're not just magically "gone".

PR #182, opened 2024-05-17
#179 Store connections in IndexedDB

nevendyulgerov:


brunomenezes (changes requested):

apps/web/src/providers/connectionConfig/connectionConfigProvider.tsx
```diff
@@ -41,6 +51,7 @@ const ConnectionConfigProvider: FC<ConnectionConfigProviderProps> = ({
                 opened={state.showConnectionModal}
                 onClose={closeModal}
                 title="Create App Connection"
+                closeOnClickOutside={false}
```
nevendyulgerov, line 54: I disabled the clone-on-click-outside behaviour to prevent closing accidentally the modal while filling connection address/url.

apps/web/src/providers/connectionConfig/indexedDbRepository.ts
```diff
@@ -0,0 +1,129 @@
...
+    async remove(addr: Address) {
+        try {
+            await this.connections
+                .where("network")
+                .equals(networkId)
+                .and((connection) => connection.address === addr)
+                .delete();
+            return true;
+        } catch (err) {
+            return false;
```
brunomenezes, line 105: Hi @nevendyulgerov, I would not catch and swallow the error here because it is up to who is calling the `remove` method to catch and decide what to do with it. That is the case for the "connection-hooks" and the `lifecycle` actions where we catch and log the error reason and also pass up in case the hook's caller attached a callback interested in listening to failures to do something in the UI (e.g. Notification).


Here for remove: https://github.com/cartesi/rollups-explorer/blob/f321db064bc47c6bfff9c6e88224c09e5546c308/apps/web/src/providers/connectionConfig/hooks.tsx#L91
Here for add: https://github.com/cartesi/rollups-explorer/blob/f321db064bc47c6bfff9c6e88224c09e5546c308/apps/web/src/providers/connectionConfig/hooks.tsx#L71

brunomenezes (changes requested):

brunomenezes (approved):

PR #197, opened 2024-06-03
#178 Create Responsive Table Component

### LINKED ISSUE #178: Create responsive table component
## 📄 Context

We already have several tables in the app and most of those need to be responsive. As the app evolves, we'll probably add more tables. Currently, the code for responsive tables is spread across their related components. It makes sense that we extract it into a single reusable component so we can:

- quickly bootstrap responsive tables in the app
- create responsive tables on the fly, without the need to remember this implementation detail (as currently is required for standard tables)

## ✔️ Solution

Create a wrapper component `ResponsiveTable` for `mantine`'s `Table` component, so that it is responsive by design. It should work as the responsive table on the inputs page.

## 📈 Subtasks

- [ ] Utilise `ResponsiveTable` in all places where responsive tables are required. 
- [ ] Remove old, duplicate code related to the responsive table behaviour.

## 🎯 Definition of Done

- [ ] All responsive tables are displayed as before but powered by the new `ResponsiveTable` component.

dandheedge:
Close #178

### DIFF
--- a/apps/web/src/components/applications/applicationRow.tsx
+++ b/apps/web/src/components/applications/applicationRow.tsx
@@ -1,4 +1,4 @@
-import { ActionIcon, Box, Group, Paper, Table, Tooltip } from "@mantine/core";
+import { ActionIcon, Box, Group, Table, Tooltip } from "@mantine/core";
 import Link from "next/link";
 import { FC } from "react";
 import {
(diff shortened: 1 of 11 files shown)


nevendyulgerov (commented):
Hey, @dandheedge , nice work on this one! I had a slightly different idea for the responsive table component though. I was thinking about implementing a component that can be invoked from jsx code like this:

```
<ResponsiveTable
  items={[{
    id: 1,
    name: 'Test abc'
  }, {
    id: 2,
    name: 'Test xyz'
  }]}
  columns={[{
    key: 'id',
    label: 'ID',
    render: (value, item, index) => (
      <span>{value}</span>
    )
  }, {
    key: 'name'
    label: 'Name',
    sticky={true}
    render: (value, item, index) => (
      <p>{value}</p>
    )
  }]}
/>
```

`items` contains the array of items (rows) that need to be displayed in the table. `columns` contains the array of column configs. Each config contains `key` (should be of type `keyof typeof item`, one of the item's keys), `label` (name of the column), `render` function that controls how data should be displayed for that row's column and some other options. 

Inside `ResponsiveTable` we can build the table columns and rows using `items` and `columns` props. Then, by passing `sticky` boolean as part of the column config the `ResponsiveTable` component can understand that this column should be sticky (like the `Data` column for the inputs table).

This implementation will hide even more code so we'll have an easier time setting up a responsive table.

Antd has a similar implementation of their [table component](https://ant.design/components/table).

Please let me know what you think @dandheedge and @brunomenezes .

nevendyulgerov:
Closing this one because I want to implement the `ResponsiveTable` component as outlined in the above comments.

Issue #186, opened 2024-05-24
Add input completion status


guidanoli:
## 📄 Context

When an input is sent to an application, it may be in one of the several possible [statuses](https://github.com/cartesi/grpc-interfaces/blob/v0.16.0/server-manager.proto#L177).
This information is already made available through the GraphQL interface, and can help developers debug their application and users have more feedback from their input.

Available Status Ref: https://github.com/cartesi/grpc-interfaces/blob/v0.16.0/server-manager.proto#L177.

## ✔️ Solution

Add the input status in the UI as a column in the table. **This would, of course, depend on a connection to a GraphQL endpoint.**

1. Add a new Column called `Status` with a interrogation sign to include a tooltip.
2. The tooltip wording is `Check the status by adding a connection. Click the ? in the row to add a connection.`
3. A row for an application without a connection should display `?` as a clickable call-to-action
 3.1. A click on it should open the `Connection creation modal`, with application address already filled.
 3.2 A mouse over it should display a tooltip with wording "Click to add a connection and inspect the input status."
4. For an application with a connection it should display the `status` informed by the app's graphQL in a `<Badge />` component just like the `Method` column.
5. Make sure the input status displayed is per index. e.g. index 1 may be `ACCEPTED` but index 2 a `REJECTED`.

## 📈 Subtasks

- [x] Add new input table header `Status` with tooltip (wording above).
- [x] Add call-to-action with hover tootip for non-connected Apps (wording above).
- [x] Add logic to handle call-to-action as stated above.
- [x] Add logic to automatic display the status of input (per input index) for connected Apps.
- [ ] Add test cases

## 🎯 Definition of Done

- [x] I can see the input's status for connected apps. 
- [x] I can hover or click the tooltip and read the intent of the new column and how to use it. 
- [x] I can click the `?` badge and start the `connection creation workflow` for that specific Application.
- [x] I can automatic see the input status for connected applications.

## Extras

<details>
<summary><h3>Screenshot</h3></summary>

![image](https://github.com/cartesi/rollups-explorer/assets/9125226/5fa062d4-a350-4204-80f7-d2b18e501a8a)
</details>

brunomenezes:
Hi @guidanoli, I see the value here. I think a column `Status` will serve better than a `tab` since that is the "state" of the content displayed by the tab (if that makes sense). Let me refine a bit here and let me know your thoughts on that. 


1. Add a new Column called `Status` with a interrogation sign to include a tooltip.
[...]

Issue #165, opened 2024-04-25
Add decode specification (v1)


brunomenezes:
## 📄 Context
Currently, for the input's payload, we are displaying the content in three ways: as is (i.e. Raw), as Text (trying to convert the Hex value) and JSON that leverages the result of the latter. It would be an excellent addition to provide a way for the user (i.e. developer) to describe how it should be parsed and save that specification for reuse.

_PS: I am calling the feature `decode specification` because the payload may be encoded with a [non-standard-packed-mode](https://docs.soliditylang.org/en/latest/abi-spec.html#non-standard-packed-mode) and just dropping an ABI will not work; pre-steps are necessary (e.g. Slice bytes) before you define a piece of ABI to use._

## ✔️ Solution
Implement a UI that allows users to specify how to decode a given input payload into a human-readable form. The spec is a single or series of transformation steps, depending on the original encoding used.

* Support JSON-ABIs format.
* Support Human-Readable-ABI format.
* Support byte slices. e.g. Portals use a non-standard packed mode [encodePacked](https://docs.soliditylang.org/en/latest/abi-spec.html#non-standard-packed-mode) so pre-steps are necessary.
* Support live feedback. i.e., set a payload while defining the specification so the user can see the result. 
* Support conditional for auto-applying the spec to a given input. e.g. `when app address x and msgSender y`
   * Limit to a couple of properties: i.e. application-address, msgSender.
   * In case of multiple specs would match. The first match is the one selected.
* Should be able to choose a named specification against a payload manually. It should transform the content inside the `Raw` panel.
* Should be able to save the specification. (local-storage)

## 📈 Subtasks

- [x] Implement the custom specification UI.
- [x] Add support to byte slicing.
- [x] Add support to JSON-ABIs.
- [x] Add support to human-readable ABI.
- [x] Add live testing feedback. 
- [x] Add support to conditionals.
- [x] Update the input details component to support specification selection.

## 🎯 Definition of Done

- [x] I can see a correct payload in human-readable form inside the `raw` panel. Also, show the specification selected.
- [x] specification can be saved. Persist through page refreshes.
- [x] Auto-select a Specification when conditional match.

brunomenezes:
cc: @tuler

brunomenezes:
[...]

PR #198, opened 2024-06-04
#186 Add input completion status

nevendyulgerov:
I added the input completion status and also covered the new logic with unit tests.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -27,6 +27,7 @@ jobs:
               env:
                   E2E_BASE_URL: ${{ github.event.deployment_status.environment_url }}
 
-            - name: Cancelling parallel jobs
-              if: failure()
-              uses: andymckay/cancel-action@0.2
+            - name: Update check status to the outcome
+              run: yarn node reportGitHubCheckStatus.mjs "End-to-end (${{ github.event.deployment_status.environment }})" ${{ github.sha }} ${{ steps.test.outcome }}
+              env:
+                GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -28,6 +28,6 @@ jobs:
                   E2E_BASE_URL: ${{ github.event.deployment_status.environment_url }}
 
             - name: Update check status to the outcome
-              run: yarn node reportGitHubCheckStatus.mjs "End-to-end (${{ github.event.deployment_status.environment }})" ${{ github.sha }} ${{ steps.test.outcome }}
+              run: yarn node reportGitHubCheckStatus.mjs "End-to-end (${{ github.event.deployment_status.environment }})" ${{ github.sha }} ${{ steps.e2e.outcome }}
               env:
                 GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -27,3 +27,7 @@ jobs:
               continue-on-error: false
               env:
                   E2E_BASE_URL: ${{ github.event.deployment_status.environment_url }}
+                  
+            - name: Cancelling parallel jobs
+              if: failure()
+              uses: andymckay/cancel-action@0.2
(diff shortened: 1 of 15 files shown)


docker/Dockerfile
```diff
@@ -5,7 +5,7 @@ RUN apk add --no-cache libc6-compat
 RUN apk update
 
 WORKDIR /app
-RUN yarn global add turbo
+RUN yarn global add turbo@1.13.3
```
nevendyulgerov, line 8: The build was failing with the following errors both from this branch and from `main`:

Screenshot 2024-06-05 at 16 12 32

Screenshot 2024-06-05 at 16 12 57

The error looked like it's coming from `turborepo` so I checked their repo and found 2 related issues:
- https://github.com/vercel/turbo/issues/6358
- https://github.com/vercel/turbo/issues/6358

Then I noticed that several new versions of turborepo have been recently published so I assumed there was a regression in those new versions that was causing this error to popup on our side.

Once I locked the version to `1.13.3` (latest stable before those recent updates) the error disappeared.

apps/web/src/components/connection/connectionInputStatusBadge.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const ConnectionInputStatusBadge: FC<ConnectionInputStatusBadgeProps> = ({
+    graphqlUrl,
+    index,
+}) => {
+    const [result] = useQuery<InputStatusQuery, InputStatusQueryVariables>({
+        context: useMemo(
+            () => ({
+                url: graphqlUrl,
+                requestPolicy: "network-only",
```
brunomenezes, line 24: Hi @nevendyulgerov, we could remove the `requestPolicy: network-only` because once the input comes to the node, and it gets this status, it is "set in stone", so it has no value making extra n calls (number of table rows) every time we change the page as the result would not change. 

The default request policy is cache-first, so it tries the cache result and only goes to the network in case of a miss.

apps/web/graphql/rollups/queries.graphql
```diff
@@ -97,4 +96,10 @@ query checkStatus {
   notices {
     totalCount
   }
+}
+
+query inputStatus($index: Int!) {
+  input(index: $index) {
```
brunomenezes, line 102: This one is just a comment, not a change request. Even though the filters in the rollups-graphql are a bit limited, I believe we could leverage the `inputs(where:{indexGreaterThan: lower_index-1, indexLowerThan: higher_index+1})` with attention to do a -1 and +1 to the bounds or they would be excluded from the result (kind of a trap). That would make the calls per page (10 or 30 items) to 1 per application of interest in that table page (The ones with connections). However, an architecture change related to UI state management will be needed to ease maintenance. I am adding State Management lib to the work on #165, so let's catch up on that. 

This improvement will come in future iterations.

brunomenezes (changes requested):

brunomenezes:
I did a rebase and pushed. 

cc: @nevendyulgerov

apps/web/src/components/connection/connectionInputStatusBadge.tsx
brunomenezes: Update: I was talking to Guilherme about that, and he asked if `unprocessed` could not be updated to `accepted` as an example. So, I sent a message to the Node-ref team to check the behaviour with accurate information.

nevendyulgerov: Thanks for those insights @brunomenezes ! I'll make the change as you suggested for now and let's see what the Node-ref team will say.

brunomenezes: Just to finish the loop. It starts as unprocessed and then moves to one of the available statuses. So, I think we can work with default behaviour rather than always going for the network since it is an update that happens only once.

brunomenezes (dismissed):

.github/workflows/e2e.yml
```diff
@@ -24,5 +24,12 @@ jobs:
 
             - name: Run e2e tests
               run: yarn test:e2e
+              id: e2e
+              continue-on-error: true
               env:
                   E2E_BASE_URL: ${{ github.event.deployment_status.environment_url }}
+
+            - name: Update check status to the outcome
```
nevendyulgerov, line 32: Hi, @brunomenezes , I regards to the failing e2e tests, I was expecting the build to fail with the related e2e job as well, but that was not the case. It was passing ok although the e2e job that we run on sepolia preview url was failing.

Because of this I had to implement this step. It will run the `reportGitHubStatusCheck` script, which will then post a status to the build. That status can be either `success` or `failure` so in the case of the latter the build will fail as well.

brunomenezes (approved):

Issue #193, opened 2024-05-30
Add Integration with Github actions to run against the live website.


nevendyulgerov:
## 📄 Context

Add a step for the e2e tests in the github workflow so that on each build the tests are run as well.

## 🎯 Definition of Done

- [ ] The build steps work as before and a new step - for running the e2e tests, is also working

PR #199, opened 2024-06-06
#193 Add build step for e2e tests

nevendyulgerov:
I created a new `e2e` workflow for the e2e tests. It runs on successful deployment and is triggered only if the deployment is for the `rollups-explorer-sepolia` environment. This way, we can use the produced vercel preview url as base url for the e2e tests. 

I had to use Protection Bypass for Automation from the vercel settings, otherwise the e2e tests were failing because they were hitting the vercel login page.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,12 +21,6 @@ jobs:
             - name: Install dependencies
               run: yarn install --frozen-lockfile
 
-            - name: Install Playwright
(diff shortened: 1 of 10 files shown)


nevendyulgerov:
Hi @brunomenezes , I think I managed to achieve what we discussed on our last meeting  - to use the vercel preview url for the e2e tests.

brunomenezes (approved):

Issue #220, opened 2024-07-30
Update mantine packages


nevendyulgerov:
## 📄 Context

[Mantine](https://www.npmjs.com/package/@mantine/core) has upgraded their library a few minor versions above what we current install in the app. We have `7.7.1` and the latest version is `7.11.2` so lets upgrade to the latest version. 

As a prerequisite check their [release logs](https://github.com/mantinedev/mantine/releases) for any potential breaking changes and also check their [repo](https://github.com/mantinedev/mantine/issues) for any raised issues related to the new version.

Make sure to throughly test the UI for any regressions in the mantine components that we use. Specifically, pay attention to the transaction forms.

## 📈 Subtasks

- [x] Check Mantine release log and github issues
- [x] Upgrade Mantine libraries to latest version 
- [x] Test the app for regressions paying special attention to the transaction forms
- [x] Fix any potential issues related to breaking changes in the new version

PR #223, opened 2024-07-31
#220 Upgrade mantine packages

nevendyulgerov:
I upgraded all `mantine` packages to their latest versions in all apps/packages that use them and tested for regressions.

In regards to the `@mantine/forms` package, I noticed a runtime error on the `/specifications/new` page that was causing the page to break. The issue was coming from the `form.setFieldValue` function being passed as a dependency to other hooks. Removing it from the dependency array solved that issue. More on this - [here](https://github.com/mantinedev/mantine/issues/6120).

Another thing I noticed was issues with some forms using `setValues` function. This function was causing the validation to be reset for the form resulting in error messages disappearing unexpectedly. Some unit tests were failing because of this. Switching to `setFieldValue` solved that issue.


brunomenezes:
@nevendyulgerov , the PR need a rebase. It may need to address other parts for the `Forms` changes since new features were added using the Mantine form.

brunomenezes:
Hi @nevendyulgerov, I think we should keep the validation on changes for all the fields regarding Specifications, but mainly, for example, the `HumanReadableABI` one, where it is important to display any `Abitype` validation failure in the signatures while defining them or right after a `paste`. 

I did a rollback on the fields + the `SpecificationFormView` and had a run it all pass. What problem did you find?

apps/web/src/components/specification/form/SpecificationForm.tsx
```diff
@@ -86,10 +86,15 @@ export const SpecificationForm: FC<SpecificationFormProps> = ({
 
     const initialValRef = useRef<SpecFormValues | null>(null);
     const { mode } = transformedValues;
-    const { setFieldValue } = form;
 
-    if (transformedValues.formMode === "EDITION" && !initialValRef.current)
+    const setFieldValue = useCallback((name: string, value: unknown) => {
```
nevendyulgerov, line 90: I had to cache this function, otherwise it was breaking the specifications page due to an infinite state update loop.

packages/ui/src/ERC1155DepositForm/DepositFormSingle.tsx
```diff
@@ -307,7 +307,8 @@ const DepositFormSingle: FC<Props> = (props) => {
                                 "erc1155Address",
                                 formattedValue,
                             );
-                            form.setValues({ amount: "", tokenId: "" });
+                            form.setFieldValue("amount", "");
```
nevendyulgerov, line 310: Using `setValues` here was causing the errors to disappear like so:

https://github.com/user-attachments/assets/53f464a5-dfc3-424b-ae98-cca039463ccf

Switching to `setFieldValue` mitigates the issue:
https://github.com/user-attachments/assets/0216f1a9-4387-4012-a84e-3a621c6b89f4

brunomenezes (approved):
Nice! It is sad that a few regressions were introduced. I believe on version 7.8.x is where most were added. Before, `setValues()` would have the same result as `Object.assign(target, source)`. Nonetheless, I think the changes are better by addressing only the bits causing troubles rather than changing the validation behaviour on every form.

PR #224, opened 2024-08-02
Feat: Add EtherPortal Specification

brunomenezes:
#### Summary
I have added the missing Ether Portal specification. Code change covers a new test case for the decoding engine with a real sample from EtherPortal. 

**Reviewers, if happy and upon approval, please feel free to merge it as it will positively reflect on automatically decoding EtherPortal-related inputs.**

PS: No changes to the underlying decoding logic.


nevendyulgerov (approved):
LGTM!

Issue #229, opened 2024-08-08
Add Export and Import for Specifications


brunomenezes:
## 📄 Context
We have a new Specification feature to help decode abi-encoded data on the input's payload. The created specs persist on local storage and belong to the domain where you run the web application (e.g., localhost). So, to avoid the tedious task of recreating them every time you run the App in a different domain, an import/export feature will be helpful.

## ✔️ Solution
Create a **versioned** importer/exporter. The exporter should generate a JSON file with specific properties (described below). The importer should be able to accept only JSON extensions and work on the designated version. 

* The **export** action should be visible when at least one specification is created. 
* Exported file name to be `cartesiscan_specifications.json`
* All specifications should be exported. Filtering does not affect the exported result (i.e. filter by JSON ABI / ABI Params). 
* The **import** action should be available when no specifications are created (adjust the existing messages to accommodate the import). It should also be available when there are one or more specifications.
* Handle all possible points of error when reading the imported file and notify the user as such. e.g., the exported version does not match, meaning the file was tampered with.
* Limit the possible file extensions to be imported to only `.json`; 

## Example of Data Format

```ts
interface SpecificationExport {
 name: 'cartesiscan_specifications_export';
 version: number;
 timestamp: number;
 specifications: Specification[]
}
```

### References: 
**Web APIs**: URL , Blob, File.

## 📈 Subtasks

- [x] Implement Exporter Logic and UI
- [x] Implement Importer Logic and UI
- [x] Handle and notify Errors .
- [x] Add tests. 

## 🎯 Definition of Done

- [x] I can successfully export my existing specifications.
- [x] I can import and use my specifications. 
- [x] I can see any errors related to the file I am trying to import.

PR #232, opened 2024-08-19
#229 Add import and export for specifications

nevendyulgerov:
I added the versioned functionality for exporting and importing specifications and covered it with unit tests.


.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
             - name: Setup Node.js environment
               uses: actions/setup-node@v3
               with:
-                  node-version: latest
+                  node-version: 22.6.0
```
nevendyulgerov, line 18: Fix for https://github.com/vercel/next.js/issues/69150. Same issue happens with our build as well.

apps/web/src/components/specification/transfer/validators/v1.ts
```diff
@@ -0,0 +1,192 @@
...
+                            ];
+                        },
+                        [],
+                    );
+
+                    let sliceErrors: (string | null)[] = [];
+                    if (specification.mode === "abi_params") {
+                        const sliceTargetError = (
+                            specification?.sliceInstructions ?? []
+                        ).some((s) => s.name === specification.sliceTarget)
```
brunomenezes, line 105: The [sliceTarget](https://github.com/cartesi/rollups-explorer/blob/main/apps/web/src/components/specification/types.ts#L66) is optional even when `sliceInstructions` are present because they're meant for non-standard encoding i.e. `abi.encodePacked`. Therefore, enforcing at least one slice's name that matches the `slice-target` defined will cause problems.

A few test cases to check what I am talking about 

* Case without `sliceTarget` [here](https://github.com/cartesi/rollups-explorer/blob/main/apps/web/test/components/specification/decoder.test.ts#L101)
* Case with `sliceTarget`[here](https://github.com/cartesi/rollups-explorer/blob/main/apps/web/test/components/specification/decoder.test.ts#L146)

apps/web/src/components/specification/transfer/validators/v1.ts
```diff
@@ -0,0 +1,192 @@
...
+                                ) => {
+                                    const error =
+                                        isNumber(sliceInstruction.from) &&
+                                        isNumber(sliceInstruction.to) &&
+                                        isNonEmptyString(
+                                            sliceInstruction.name,
+                                        ) &&
+                                        isNonEmptyString(sliceInstruction.type)
+                                            ? null
+                                            : "Invalid slice schema.";
```
brunomenezes, line 123: Check the [ SliceInstruction type here](https://github.com/cartesi/rollups-explorer/blob/main/apps/web/src/components/specification/types.ts#L40), but the `[name, to , type]` properties are not required even though the `name` is required in the form. Check [this test case ](https://github.com/cartesi/rollups-explorer/blob/main/apps/web/test/components/specification/decoder.test.ts#L146) or the `systemSpecs.ts` file for reference. So importing a good file missing such fields will fail to proceed.

apps/web/src/components/specification/transfer/validators/v1.ts
```diff
@@ -0,0 +1,192 @@
...
+                                      .map((o) => o.value)
+                                      .join(", ")}`;
+
+                            const schemaErrors = condition.conditions.reduce(
+                                (
+                                    conditionAccumulator: (string | null)[],
+                                    conditionItem,
+                                ) => {
+                                    const error =
+                                        isNonEmptyString(conditionItem.field) &&
```
brunomenezes, line 77: As you are also doing a run in the `operators` list below. I think you could check the [inputProperties](https://github.com/cartesi/rollups-explorer/blob/main/apps/web/src/components/specification/types.ts#L15) for the `field` property. A rebase with the main is necessary since we merged the #233, and patching is necessary when comparing for retro compatibility's sake. There is a function that does that [here](https://github.com/cartesi/rollups-explorer/blob/main/apps/web/src/components/specification/conditionals.ts#L66) , it was added in the PR mentioned.

apps/web/test/components/specification/stubs.ts
```diff
@@ -112,3 +117,80 @@ export const inputResponses = {
     nonPortalRelatedInput,
     dappAddressRelayInput,
 } as const;
+
+/**
+ * Specifications export
+ */
+export const defaultSpecificationExport: SpecificationTransfer = {
+    version: 1,
+    timestamp: new Date().getTime(),
```
brunomenezes, line 126: This is just a warning. Checking that property in a "deterministic" way may break your test.

brunomenezes (changes requested):
@nevendyulgerov **A rebase is necessary**. I added comments related to the validations that are happening. Also, I got stuck after exporting and trying to re-import my existing specifications.

#### Steps
* Export existing Specifications 
* Then, delete all of them. 
* After trying to import and seeing an error message related to `slice target` that does not exist in my specifications. 
* Nothing happens when I close the error and try to re-import the same file. It is like the button is not working; neither could I see the error again, nor were the specifications loaded.

#### Exported JSON file
[cartesiscan_specifications_export.json](https://github.com/user-attachments/files/16831689/cartesiscan_specifications_export.json)

brunomenezes:
@nevendyulgerov I opened a PR against this branch-PR #242. Because I noticed a problem with Specifications showing up, and after refreshing the page, they're gone (recording below). I would suggest the import/export button be the default variant so it follows all the other buttons' look and feel (e.g. `Connect wallet`, `New` etc).

brunomenezes (approved):

Issue #234, opened 2024-08-26
Update vitest libraries


nevendyulgerov:
## 📄 Context

[Vitest](https://www.npmjs.com/package/vitest) has upgraded their framework two major versions above what we current install in the app. We have `0.34.2` and the latest version is `2.5.0` so lets upgrade to the latest version. 

As a prerequisite check their [release logs](https://github.com/vitest-dev/vitest/releases) for any potential breaking changes and also check their [repo](https://github.com/vitest-dev/vitest) for any raised issues related to the new version.

## 📈 Subtasks

- [x] Check Vitest library release log and github issues
- [x] Upgrade the library to latest version 
- [x] Ensure all existing unit tests work as before and pass ok

PR #235, opened 2024-08-26
#234 Upgrade Vitest libraries to latest version

nevendyulgerov:
TBD.


apps/web/vitest.config.mts
```diff
@@ -4,9 +4,9 @@ import { defineConfig, UserConfig } from "vitest/config";
 export default defineConfig({
     plugins: [react()],
     test: {
-        exclude: ["e2e/**"],
+        exclude: ["e2e/**", "node_modules/**"],
```
nevendyulgerov, line 7: I noticed that the `vitest` command runs tests in `node_modules` if any exist, that's why I'm excluding this folder here.

apps/web/vitest.config.mts
```diff
@@ -4,13 +4,20 @@ import { defineConfig, UserConfig } from "vitest/config";
...
         globals: true,
-        environment: "jsdom",
+        environment: "happy-dom",
         setupFiles: "vitest-setup.ts",
         coverage: {
             reporter: ["text", "lcov"],
-            exclude: ["**/graphql/**"],
+            exclude: [
```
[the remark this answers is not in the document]
brunomenezes: Hey @nevendyulgerov, doesn't a `include: ['**/src']` help with that? I have an API PR opened, and the same changes were done [Vitest.config.mts](https://github.com/cartesi/rollups-explorer-api/blob/41-multichain-indexing/vitest.config.mts)

apps/web/package.json
```diff
@@ -64,25 +65,27 @@
...
+        "@testing-library/jest-dom": "^6.5.0",
         "@testing-library/react": "^16.0.0",
         "@types/node": "^20",
         "@types/ramda": "^0.30.1",
         "@types/react": "^18.3.3",
         "@types/react-dom": "^18.3.0",
         "@vitejs/plugin-react": "^4.1.0",
-        "@vitest/coverage-v8": "0.34.2",
+        "@vitest/coverage-v8": "^2.0.5",
+        "@vitest/ui": "^2.0.5",
```
nevendyulgerov, line 76: I added this package so that we can use the [Vitest UI](https://vitest.dev/guide/ui) if we want with the `yarn test:ui` script.

apps/web/test/components/connection/connectionForm.test.tsx
```diff
@@ -58,7 +58,7 @@ describe("connectionForm", () => {
         ).toBeInTheDocument();
 
         expect(screen.getByTestId("icon-test-inactive")).toBeInTheDocument();
-        expect(screen.getByText("Save")).toBeInTheDocument();
+        expect(screen.getByTestId("connection-save")).toBeInTheDocument();
```
nevendyulgerov, line 61: More tests started failing after the latest rebase. Probably something in `@testing-library/*` or `@mantine/*` changed in a patch version, but I haven't dig into this further.

nevendyulgerov:
Hi, @brunomenezes , I checked the issue with the reduced test coverage. 

From the looks of it, we are covering the same files with the inclusion of `.ts` files that contain only type definitions - interfaces, types, enums, etc. [Example](https://coveralls.io/builds/69554992/source?filename=packages%2Fui%2Fsrc%2FDepositFormTypes.ts).

For those files the reporter is giving a 100% coverage though, so they are not bringing down the coverage.

I compared the same file 

- https://coveralls.io/builds/69554992/source?filename=packages%2Fui%2Fsrc%2FTransactionProgress.tsx

- https://coveralls.io/builds/69558945/source?filename=packages%2Fui%2Fsrc%2FTransactionProgress.tsx#L8

between this branch and the `main` branch. Here's a screenshot:

Screenshot 2024-09-04 at 14 55 57

The only difference I noticed was that the individual imports are not marked with `✔` anymore. The remaining content of both files is identically covered . So maybe the difference in the imports coverage is causing the percentage to go down but I'm not sure. In any case, it appears that the reporter is giving different results for the same file contents.

I also tried a different reporter - [@vitest/coverage-istanbul](https://www.npmjs.com/package/@vitest/coverage-istanbul) but with it, the coverage fell even more - https://coveralls.io/builds/69600513.

So, I think can either ignore the failed coveralls check for this PR and merge it like this with this new test coverage results or I can add more unit tests to prevent the decrease of test coverage.

brunomenezes:
> Hi, @brunomenezes , I checked the issue with the reduced test coverage.
> 
> From the looks of it, we are covering the same files with the inclusion of `.ts` files that contain only type definitions - interfaces, types, enums, etc. [Example](https://coveralls.io/builds/69554992/source?filename=packages%2Fui%2Fsrc%2FDepositFormTypes.ts).
> 
> For those files the reporter is giving a 100% coverage though, so they are not bringing down the coverage.
> 
> Between this branch and the `main` branch I compared the same file (`TransactionProgress`):
> 
> * https://coveralls.io/builds/69554992/source?filename=packages%2Fui%2Fsrc%2FTransactionProgress.tsx
> * https://coveralls.io/builds/69558945/source?filename=packages%2Fui%2Fsrc%2FTransactionProgress.tsx#L8
> 
> Here's a screenshot:
> 
> Screenshot 2024-09-04 at 14 55 57
> 
> The only difference I noticed was that the individual imports are not marked with `✔` anymore. The remaining content of both files is identically covered. Also, both files have the same content. So it appears that the difference in the imports coverage is causing the percentage to go down. A confirmation for this is the fact that the number of uncovered lines in `TransactionProgress` (43) equals the number of lines for which there is no more a `✔` symbol:
> 
> Screenshot 2024-09-04 at 17 08 16
> 
> So, it appears that the reporter is giving different results for the same file contents.
> 
> I also tried a different reporter - [@vitest/coverage-istanbul](https://www.npmjs.com/package/@vitest/coverage-istanbul) but with it, the coverage fell even more - https://coveralls.io/builds/69600513. Additionally, I checked vitest's repo for any related issues. I found this [one](https://github.com/vitest-dev/vitest/pull/5457) and applied the suggested solution with `ignoreEmptyLines` but this didn't affect the coverage result.
> 
> So, I think we can either ignore the failed coveralls check for this PR and merge it like this with this new test coverage result or I can add more unit tests to prevent the decrease of test coverage.

@nevendyulgerov, that is alright. As we discussed, they changed a bit about how mapping works. That becomes the new threshold, and we will work from here.

brunomenezes (approved):

Issue #205, opened 2024-07-01
Add e2e utility function for logging into Metamask


nevendyulgerov:
## 📄 Context

Add an e2e utility function for logging into MetaMask. We'll need this function in all other e2e tests related to the Send Transaction UI.

## 🎯 Definition of Done

- [x] The new e2e utility function allows you to log into MetaMask with a test account.

nevendyulgerov:
I did some work for this one in https://github.com/cartesi/rollups-explorer/tree/feature/205-add-e2e-utility-function-for-logging-into-metamask, exploring the [synpress](https://www.npmjs.com/package/@synthetixio/synpress) library. For the time being I'll leave this as is and will get back to it when I'm idle.

PR #240, opened 2024-09-11
#205 Add e2e utility for logging into MetaMask

nevendyulgerov:
Using [@synthetixio/synpress](https://www.npmjs.com/package/@synthetixio/synpress) I added a MetaMask fixture for the e2e tests and I also added a simple test for validating that we can log in MetaMask. Here's a video of it for reference:

https://github.com/user-attachments/assets/18ea045b-286c-4d42-9c26-61be029d5e08


apps/web/additional.d.ts
```diff
@@ -36,3 +36,7 @@ declare namespace NodeJS {
         VERCEL_AUTOMATION_BYPASS_SECRET?: string;
     }
 }
+
+declare module "@synthetixio/synpress/commands/metamask";
```
nevendyulgerov, line 40: The library lacks TS support, it doesn't have `@types/*` package either so that's why I declared modules for it here.

apps/web/e2e/fixtures/metamask.ts
```diff
@@ -0,0 +1,56 @@
...
+            args: browserArgs,
+        });
+
+        // Wait for Metamask window to be shown.
+        await context.pages()[0].waitForTimeout(3000);
+
+        // Setup metamask
+        await initialSetup(chromium, {
+            secretWordsOrPrivateKey:
+                "test test test test test test test test test test test junk",
```
nevendyulgerov, line 47: Looking at the internal logic of `initialSetup` and `secretWordsOrPrivateKey` it seems that we can either import a wallet or create a new one and import an account:

```
if (secretWordsOrPrivateKey.includes(' ')) {
  // secret words
  await module.exports.importWallet(secretWordsOrPrivateKey, password);
} else {
  // private key
  await module.exports.createWallet(password);
  await module.exports.importAccount(secretWordsOrPrivateKey);
}
```

I also see a `switchAccount` function available from `metamask`.

I can explore those further to see how we can hook an existing account with SepoliaETH in it.

apps/web/package.json
```diff
@@ -63,8 +63,9 @@
         "@graphql-codegen/typescript-operations": "^4",
         "@graphql-codegen/typescript-urql": "^4",
         "@graphql-typed-document-node/core": "^3",
-        "@playwright/test": "^1.43.1",
+        "@playwright/test": "^1.37.0",
```
nevendyulgerov, line 66: I had to downgrade `@playwright/test` because `@synthetixio/synpress` depends on `1.37.0`.

.github/workflows/e2e.yml
```diff
@@ -6,15 +6,17 @@ jobs:
         if: github.event_name == 'deployment_status' && github.event.deployment_status.state == 'success' && contains(github.event.deployment_status.environment_url, 'rollups-explorer-sepolia')
         timeout-minutes: 30
         runs-on: ubuntu-latest
+        container:
```
nevendyulgerov, line 9: I added this container as per instructions [here](https://playwright.dev/docs/ci#via-containers). With it, we can run `xvfb-run` out-of-the-box. We need `xvfb` for desktop emulation in regards to the metamask browser plugin.

apps/web/e2e/transactions/depositEther.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { expect } from "@playwright/test";
+import { test } from "../fixtures/metamask";
+import * as metamask from "@synthetixio/synpress/commands/metamask";
+
+test.describe.configure({
+    mode: "serial",
+    timeout: 120000,
```
nevendyulgerov, line 7: I noticed that the metamask context takes a while to be set up - `await chromium.launchPersistentContext` is rather slow, it was failing with a 1 minute timeout. That's why I increased it to 2 minutes.

apps/web/package.json
```diff
@@ -13,7 +13,7 @@
         "test:watch": "vitest",
         "test:ui": "vitest --ui",
         "test:ci": "vitest run --coverage",
-        "test:e2e": "playwright test",
+        "test:e2e": "playwright test ./e2e/layout/ ./e2e/pages && xvfb-run npx playwright test ./e2e/transactions/",
```
nevendyulgerov, line 16: I had to split the tests in 2 groups because the metamask tests were failing due to the context not being setup (even with a 2 minute timeout):

1. Normal e2e tests that do not require metamask setup
2. Transaction e2e tests that require metamask setup

If the first group fails the second won't be executed.

brunomenezes (approved):

PR #246, opened 2024-09-20
feat: Voucher payload auto decoding

brunomenezes:
### Summary
Code changes to support auto-decoding the voucher payload. As the voucher contains information on the destination (contract address) and the payload (4-byte selector + arguments), we request the destination's ABI and try to decode it using our Specification on the fly. **In any failures while hopping into the network or during the decoding attempt, it will gracefully continue to work by just returning and rendering the original raw payload (i.e. Hex).** Other changes and refactorings were done along the way. 

#### Changes summary:

* apps/web
  * Added a new hook to handle voucher payload automatically where possible.
  * Refactored the function to get the configured chain ID to be self-contained.
  * Improved meta for decoded payloads with `JSON-ABI` spec type. 
  * Improve error handling in the `VoucherExecution` component.
    * The call to the `CartesiDapp#wasExecuted` method; Now, the `chainId` prop is set instead of relying on underneath logic. That avoids unnecessary error notifications when switching between CartesiScan networks and the wallet is disconnected. 
    * It will only start the **simulated** call once the contract call to check whether a voucher was executed has a response and the voucher has its proof.
    * Changed the text's precedence for the execution button. 
* packages/ui
  * Refactored the `InputDetails -> Content` component to fix an internal react error. It loses access to an expected static flag, which causes odd behaviours: it re-renders the previous state of the component (as a perceived fallback).
  * Modified how to identify what type of `PageableContent`it is by passing a param value rather than relying on **static function property** (PageableContent is shared by the output types, i.e. Notices / Reports / Vouchers)


packages/ui/src/InputDetails/Content.tsx
```diff
@@ -42,54 +49,44 @@ export const ContentTypeControl = ({
...
+            ? hexToString(content)
+            : content;
+    const ref = useRef<HTMLTextAreaElement | null>(null);
 
-            useEffect(() => {
-                if (ref.current !== null) {
-                    ref.current.blur();
-                }
-            }, [value]);
+    useEffect(() => {
```
nevendyulgerov, line 64: This effect will be invoked on each render, if this is desired, we can remove the effect altogether and just run the code at the top-level of that component. Otherwise, we should include an empty array for its dependencies.

brunomenezes: My first option is always not to use a `useEffect`. The goal is to have a `<JsonInput>` content formatted, and it does not work when the blur is triggered in the top level. Thus, a "microtask" must run after the content is visible and in-focus.

Issue #251, opened 2024-09-25
Add ABI encoding capability to `send raw input`


brunomenezes:
## 📄 Context
Currently, when sending a transaction, the `Raw input` form provides two ways of filling the data to be sent: `Raw`, i.e., as a Hex value, or `String to Hex`, where the user sets a string value, and the application will convert it to Hex. An interesting third way is for the user to define or choose an ABI, decide the function method and set the argument(s) value(s) if any is required. The application would take care of the ABI encoding the data. 

**Also, it is an excellent moment to rebrand that form name from `Raw Input` to `Generic Input`.**

## ✔️ Solution

Change the form name to reflect its capabilities and add the ABI encoding feature with the following requirements:

* Rebrand the `Raw Input` form to `Generic Input` form
* The user should be able to select an existing Specification (JSON_ABI type only)
* The user should be able to define an ABI on the fly (human-readable format or JSON ABI format)
* Once the ABI is available to the form (filled or selected):
    * A `combobox` with all the ABI items where the type is `function` should be listed (i.e. no event, constructor, etc, to be included).
    * The text content should be the function signature, e.g. `ownerOf(uint256)`.
* Once a function is selected and it has arguments; 
    *  Dynamic inputs should become available following the ABI item inputs order, with the name as the input's label (fallback to the ABI item input type in case the name is missing, e.g. address, uint, string). 
    * The order of these dynamic inputs must be respected so the transaction is prepared correctly. 
* If a selected function has no arguments;
    * Display a simple centralised info level message using copy “No input arguments defined for {#function signature}.”
* The send should be enabled once all required fields are filled out and no errors are captured.

**PS: Check [Viem docs](https://viem.sh/docs/getting-started) to support conversions here and there e.g. function signatures**

## 📈 Subtasks

- [x] Rename the form name from `Raw Input`.
- [x] Implement the new ABI encoding logic.
- [x] Handle errors.
- [x] Add tests.

## 🎯 Definition of Done

- [x] I can ABI encode my data and send a transaction to my Cartesi Application.
- [x] I can receive the ABI encoded data on my Cartesi Application and decode it correctly using the same ABI. 
- [x] CI checks are passing.
- [x] Vercel deployments are live with success.

PR #256, opened 2024-10-14
#251 Add ABI encoding for raw input form

nevendyulgerov:
I rebranded the form and reorganised related files for it under `packages/ui/src/GenericInputForm`.

Summary of the changes:
- Changed form name to `Generic Input`
- Added `ABI to Hex` mode and slightly changed labels for the other modes to resemble those from `cartesi-cli`
- Added fields for the `ABI to Hex` mode as follows:
  - `ABI method` select for choosing whether to use existing specifications or create a new one
    - When the existing specification option is selected:
      - Select with options for choosing an existing specification is displayed
      - Select with options for choosing a function from that existing specification is displayed
      - Text fields for the parameters for the selected function are displayed
    - When the new specification option is selected:
      - Segmented control is displayed with 2 options - `JSON ABI` and `ABI Parameters`:
        - When `JSON ABI` is selected:
          - Textarea for the human ABI is displayed
          - Select with options for choosing a function from the filled human ABI is displayed
          - Text fields for the parameters for the selected function are displayed
        - When `ABI Parameters` is selected: 
          - Text field for the ABI params is displayed
          - Text fields for the parameters from the filled ABI params are displayed
- For encoding the parameters, I'm using `encodeAbiParameters` from `viem`.
- I used as a reference for the implementation `cartesi-cli` and more specifically https://github.com/cartesi/cli/blob/main/apps/cli/src/commands/send/generic.ts


brunomenezes (changes requested):
Hi @nevendyulgerov, 
There is a problem when ABI encoding uses a `JSON_ABI` format. I notice in the [index.ts#encodeFunctionParamsDebounced()](https://github.com/cartesi/rollups-explorer/blob/f7ba14d85742e9f951e436b91d75edb07d98444e/packages/ui/src/GenericInputForm/index.tsx#L79) is doing the `encoding`, but the underneath function assumes it is an `ABI parameter` (i.e. on-the-fly definition), but that is not what we want for `JSON_ABI` because it will not include the 4-bytes selector that is the function signature for the `JSON_ABI` format; Therefore, both the DApp backend will fail to parse its payload and also CartesiScan will fail to do the same using the JSON_ABI specification saved. 


I'll use the onChess ABI as an example:
**network deployed: 84532 (base-sepolia)**
**Address: 0xc93796ff6ed6b8d15d68ecb793df221ecf042774**

Getting the input index 13, we have an input encoded by the function `move` that expects two arguments (address game, string move)

Suppose I follow the same steps through the form. I expect the same encoded value to be generated as the `rawInput` that goes to the `InputBox`. But it generates a different output.


Expected: `0x17799d270000000000000000000000002189f5ec6516583b6a0cddf47c712355316d0ff6000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000026436000000000000000000000000000000000000000000000000000000000000`

Generated: `0x0000000000000000000000002189f5ec6516583b6a0cddf47c712355316d0ff6000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000026436000000000000000000000000000000000000000000000000000000000000`

It is just the function that needs to change based on the ABI definition chosen (ABI Parameters or JSON-ABI). For the latter the 4-bytes selector needs to be included because it tells the decoding function what ABI's function to pick. 

I added information about the onChess app + the ABI so you can play around.  Reach out if you have questions.

<details>
 <summary><h3>onChess ABI</h3> </summary>
 
  ```solidity
function create(uint256 bet, string timeControl, uint32 minRating, uint32 maxRating) returns (address)
function cancel()
function move(address game, string move)
function resign(address game)
function claim(address game)
function withdraw(uint256 amount)
function withdrawRake()
function setRakeDivider(uint32 rake)
function transferOwnership(address newOwner)
function upgrade(address dapp)
 ```
 
 ```json
 [
  {
    "name": "create",
    "type": "function",
    "stateMutability": "nonpayable",
    "inputs": [
      {
        "type": "uint256",
        "name": "bet"
      },
      {
        "type": "string",
        "name": "timeControl"
      },
      {
        "type": "uint32",
        "name": "minRating"
      },
      {
        "type": "uint32",
        "name": "maxRating"
      }
    ],
    "outputs": [
      {
        "type": "address"
      }
    ]
  },
  {
    "name": "cancel",
    "type": "function",
    "stateMutability": "nonpayable",
    "inputs": [],
    "outputs": []
  },
  {
    "name": "move",
    "type": "function",
    "stateMutability": "nonpayable",
    "inputs": [
      {
        "type": "address",
        "name": "game"
      },
      {
        "type": "string",
        "name": "move"
      }
    ],
    "outputs": []
  },
  {
    "name": "resign",
    "type": "function",
    "stateMutability": "nonpayable",
    "inputs": [
      {
        "type": "address",
        "name": "game"
      }
    ],
    "outputs": []
  },
  {
    "name": "claim",
    "type": "function",
    "stateMutability": "nonpayable",
    "inputs": [
      {
        "type": "address",
        "name": "game"
      }
    ],
    "outputs": []
  },
  {
    "name": "withdraw",
    "type": "function",
    "stateMutability": "nonpayable",
    "inputs": [
      {
        "type": "uint256",
        "name": "amount"
      }
    ],
    "outputs": []
  },
  {
    "name": "withdrawRake",
    "type": "function",
    "stateMutability": "nonpayable",
    "inputs": [],
    "outputs": []
  },
  {
    "name": "setRakeDivider",
    "type": "function",
    "stateMutability": "nonpayable",
    "inputs": [
      {
        "type": "uint32",
        "name": "rake"
      }
    ],
    "outputs": []
  },
  {
    "name": "transferOwnership",
    "type": "function",
    "stateMutability": "nonpayable",
    "inputs": [
      {
        "type": "address",
        "name": "newOwner"
      }
    ],
    "outputs": []
  },
  {
    "name": "upgrade",
    "type": "function",
    "stateMutability": "nonpayable",
    "inputs": [
      {
        "type": "address",
        "name": "dapp"
      }
    ],
    "outputs": []
  }
]
 ```
 
 </details>

brunomenezes (changes requested):
Hey @nevendyulgerov, I think that is it regarding swapping the decoding function. I got a last thing during my tests, and it is related to Tuples as function params (by defining a struct or by inline definition). The UI does not support that parameter type, as shown in the screenshot below. In that example, it only says the type and displays one field. But for tuples, the UI would need to iterate over the `components` to build the inputs for that tuple with the same rules, i.e. respecting the order. Keep in mind that a tuple can have other tuple definitions inside. It is a complex object rather than a single value. I am also adding a definition for you to play around with in the specifications for JSON_ABI so you can see the structure generated. Still, as you can define it inline, it also applies to ABI parameters.

<details>
<summary>Screenshots</summary>

image

</details>

<details>
<summary>Human-Readable ABI</summary>

```solidity
struct User { string name; uint amount; string role}
function createAccount(User user)
function createAccount2((string name, uint amount, address) user, address from)
```

</details>

brunomenezes (changes requested):
Hi @nevendyulgerov, I had a few problems while testing the changes. I will leave here the JSON_ABI specification I created to test the generated form so you can replicate the situation. Two of them became interconnected, leaking between functions and ABIs, and the other is not related to the form generation. I will also include gifs for visual ref. I could not finish the test as the `Hex` was never generated.

1. The application address is prematurely cleaned up when I change the way to encode the data. (i.e. Choosing between `Hex`, `String to Hex` and `ABI to Hex`). The format should not be related to the App address, as it is the target to which I will send the input. 
2. When using the ABI definition below, pick the `deposit` function that contains a tuple followed by two parameters in the first level. The form is generated nicely, but when adding the second and third tuple's parameters, the first-level parameters are updated (i.e. `address from`, `string reason`), and vice versa. Also, the validation messages are mixed when the function input size and tuple components size match and the types differ. However, the worst case is if the types match because the overwrite still happens (two-way data-binding), but there are no validation error messages. The form stays forever invalid as the `Send` button is permanently disabled.
3. The last one is that the validation error messages that the previous one generated are not cleaned between function selection change nor ABI change followed by selecting a new function. Replicate with the so-called `Tuples ABI` defined below to see the errors, then switch to the `others ABI` and pick the function `create` that has four parameters; the validation error messages will display automatically. 


<details>
<summary><h2>ABIs<h2></summary>

<h3>Tuples ABI</h3>

```solidity
struct User { string name; uint amount; address}
function createAccount(User user, string role)
function deposit(User user, address from, string reason)
```

<h3>onChess (in case you don't have a spec already saved)</h3>

```solidity
function create(uint256 bet, string timeControl, uint32 minRating, uint32 maxRating) returns (address)
function cancel()
function move(address game, string move)
function resign(address game)
function claim(address game)
function withdraw(uint256 amount)
function withdrawRake()
function setRakeDivider(uint32 rake)
function transferOwnership(address newOwner)
function upgrade(address dapp)
```

</details>

<details>
<summary><h2>GIFs</h2></summary>

<h3>1. problem (Application Address Cleanup)</h3>

premature-app_address_cleanup


<h3>2. Input Ref problem and mixed error messages (Not here, but this causes the forever invalid form state when types match)</h3>

inputs-ref-error-with-diff-level-overwrites


<h3>3. Leaking stale validation messages between functions and ABI changes</h3>

stale-validation-messages-leaking


</details>

brunomenezes (changes requested):
Hi @nevendyulgerov , 
I tested it, and it worked for a simple tuple + extra parameters in a function signature. Still, if I have a multi-level tuple, **it stays invalid without error messages, and I can't proceed**. The local field validations work, but nothing happened when I added the parameters. I did a quick logging, and it is returning `false` to the `form.isValid()` call. I noticed an `abiFunctionParams` array that is a `flat` version of these values, and I'm unsure how you're matching them to validate. As a comment, a recursive approach might be helpful where nesting is possible and help with final validations because the flat array is obscure regarding the matching. I believe the problem lies in this piece.

PS: Even though the ABI below has the name defined, it shall not be. So rely on something other than the name.


**Use the below ABI to replicate**

```solidity
struct Country {string code; string name;}
struct UserInfo {Country country; string name; address wallet} 
function createAccount(UserInfo user, string ref)
```

packages/ui/src/GenericInputForm/utils.ts
```diff
@@ -0,0 +1,196 @@
+import { AbiParameter, getAddress, parseAbi, parseAbiParameters } from "viem";
+import {
+    AbiInputParam,
+    AbiValueParameter,
+    FormSpecification,
+    FinalValues,
+    FormValues,
+    FormTransformedValues,
+} from "./types";
+import { prepareSignatures } from "web/src/components/specification/utils";
```
brunomenezes, line 10: This is a boundary leak in the sense that `packages` know about `apps`. A change in the `apps/web` in this function would have an unexpected effect on the `packages/ui`. That also creates an odd "circular" dependency: `web` depends on `pkg/ui`, but `pkg/ui` also depends on `web`. 

I think it would be sensible to copy the `prepareSignature` to this `utils` file and use it here and also in the `validations.ts` (I added a comment there.)

Do you agree @nevendyulgerov?

packages/ui/src/GenericInputForm/validations.ts
```diff
@@ -0,0 +1,130 @@
+import { isAddress, isHex, parseAbi, parseAbiParameters } from "viem";
+import { FormValues } from "./types";
+import { isBlank } from "ramda-adjunct";
+import { prepareSignatures } from "web/src/components/specification/utils";
```
brunomenezes, line 4: highlighted reasons [here](https://github.com/cartesi/rollups-explorer/pull/256/files#r1855693819)

packages/ui/src/GenericInputForm/utils.ts
nevendyulgerov: Good catch, @brunomenezes , I refactored it.

brunomenezes (approved):

Issue #275, opened 2024-12-10
Make the specifications preview section sticky


nevendyulgerov:
## 📄 Context

On the `/specifications/new` page, we have a preview section (and textarea) that becomes difficult to use if you have a lot of content to the left side where the form is displayed. This makes interacting with the preview challenging as you need to scroll up and down in order to use the textarea.

We can improve this by making the preview section sticky so that when users scroll down it remains within the viewport.

## 🎯 Definition of Done

- [ ] Scrolling up and down shouldn't affect the preview section visibility. It should remain visible at all times.

PR #276, opened 2024-12-10
#275 Make specs preview section sticky

nevendyulgerov:
I made the preview section sticky so that it's visible at all times.

### DIFF
--- a/apps/web/src/components/specification/components/DecodingPreview.tsx
+++ b/apps/web/src/components/specification/components/DecodingPreview.tsx
@@ -1,6 +1,15 @@
 "use client";
 import { CodeHighlight } from "@mantine/code-highlight";
-import { Alert, Card, Stack, Text, Textarea, Title } from "@mantine/core";
+import {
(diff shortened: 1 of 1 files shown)


apps/web/src/components/specification/components/DecodingPreview.tsx
```diff
@@ -15,7 +15,7 @@ export const DecodingPreview = () => {
     const content = envelope?.result ? stringifyContent(envelope.result) : null;
 
     return (
-        <Card shadow="sm" withBorder>
+        <Card shadow="sm" withBorder pos="sticky" top="4.75rem">
```
brunomenezes, line 18: Hello @nevendyulgerov, regarding the `top` attribute, what do you think about getting that number from the `theme`? It has the `components -> AppShell` and config for the header's height. Then, a sum could be done to offset and keep the spacing stable even if the height of `header` changes.

brunomenezes (approved):

Issue #281, opened 2024-12-16
Navigation issues when using input search in conjunction with pagination


nevendyulgerov:
## 🙂 Expected behavior

The url flashes and it doesn't reflect the correct search queries.

## 🫠 Actual behavior

The url should reflect the correct search queries based on the current search and the selected page and limit.

## 🧪 Minimal test case

1. Go to `/inputs` page
2. Using the top pagination, select page `2`
3. Insert some search query in the search input
4. Click the browser's Back button

## 🌎 Environment

Any environment - Mainnet / Sepolia / Local
Any rollups version.

## ✔️ Possible solutions

There are two custom hooks that control the search query params - `usePaginationParams` and `useQueryParams`. The issue is caused by these two hooks causing a race condition for the url search params. Instead of using two custom hooks, it's better to use just one where all 3 search params - `page`, `limit` and `query` are being handled.

PR #282, opened 2024-12-16
#281 Fix url search query params

nevendyulgerov:
I deprecated the custom hooks `useQueryParams` and `usePaginationParams` and unified their logic into a new custom hook - `useUrlSearchParams` where all three params (`page`, `limit`, `query`) are being handled.


apps/web/src/hooks/useUrlSearchParams.ts
```diff
@@ -12,35 +11,38 @@ export const limitBounds = {
 export type LimitBound = (typeof limitBounds)[keyof typeof limitBounds];
 
 export type UsePaginationReturn = [
```
brunomenezes, line 13: I think you forgot to rename the Interface to also match the new hook's name

cc: @nevendyulgerov

apps/web/src/components/search.tsx
```diff
@@ -1,26 +1,36 @@
...
-    const [keyword, setKeyword] = useState<string>(query);
+    const [{ limit, page, query }, updateParams] = useUrlSearchParams();
+    const [search, setSearch] = useState<string>(query);
+    const lastSearch = useRef(search);
 
     useEffect(() => {
-        updateQueryParams(keyword);
-        onChange(keyword);
-    }, [query, onChange, keyword, updateQueryParams]);
+        if (lastSearch.current !== query) {
```
brunomenezes, line 17: I wonder if that `useEffect` is necessary and if the logic ever runs because of how the `onSearch` callback is defined below. 

cc: @nevendyulgerov

nevendyulgerov: Yes, this effect helps sync the search input value with the url's `search` query param when you go back and forth in history:

https://github.com/user-attachments/assets/cf3df668-2cfa-4eb5-85c5-d744dc4031c0

brunomenezes (approved):

Issue #280, opened 2024-12-13
Ether deposit - Improve error message checks


brunomenezes:
## 📄 Context

When using the deposits through the Ether portal, the validation for the amount exceeding the balance takes a long time as it is waiting for a `prepare` call to resume. Instead, retrieving the balance information from the wallet connected and adding this new validation as the user fills the form would be faster and more efficient.

## ✔️ Solution
Get the current ETH balance of the connected account and include it in the validation for the amount the user defines rather than wait for the return of the `prepare` checks.

wording: `The amount ${amount} exceeds your current balance of ${balance} ETH`

## 📈 Subtasks

- [x] Add the new validation for faster local validation in terms of an amount exceeding the balance.
- [x] Add test cases.

## 🎯 Definition of Done

- [x] As I try to add an amount bigger than my balance, I can quickly see the error message.
- [x] An invalid form will not allow the user to proceed with the deposit.
- [x] CI checks are passing.
- [x] Vercel preview builds are passing.

PR #269, opened 2024-11-27
Feat: Support rollups v2

brunomenezes:
### Summary
Code changes to support both v1 and v2 rollup cartesi applications. 

I suggest the reviewer look commit by commit as each is related to blocks of changes that make sense and are smaller rather than the `Files changed` as they're loose and it is just too much cognitive power. **My two cents as a reviewer myself.**

Summary of changes: 

* Added new rollups-v2 contracts for wagmi code generation.
* Added new graphQL schema, queries and codegen for v2.
* Added new system Specification for the new v2 contracts (e.g. EtherPortal, ERC20Portal, etc)
* Update method resolver and how to check the input row status for v2 applications. 
* Update page navigation now, including the version in the URL. Also, handle cases where the version is not in the URL.
* Added voucher execution support for v2 apps. 
* Added Rollup v2 Output ABI to support decoding the voucher's payload from an Application v2. It has a different structure. 
* Upgrade Send transaction functionality :
   * Update EtherPortal deposit to support v1 and v2 apps.
   * Update ERC20Portal deposit to support v1 and v2 apps.
   * Update ERC721Portal deposit to support v1 and v2 apps.
   * Update ERC1155 single and batch portal deposit to support v1 and v2 apps. 
   * Update Generic Input form (a.k.a InputBox `add input`) form to support v1 and v2 apps.
   * Update the DAppAddress relay form to support only v1 apps. The search will always be filtered by v1. 
* Changed how to get human readable method and name for addresses that are related to Rollup Contracts e.g. `EtherPortal` and method `etherDeposit` that are displayed in the list of inputs.
* Refactored a search e2e test case to improve resilience https://github.com/cartesi/rollups-explorer/pull/269/commits/bb248888f2c78ae1aeb2e368487ae7bda6e4ae13. It was also breaking check the **link below**
* Skipped the remaining e2e tests that interact with Metamask (Last `main` build is actually not passing [action here](https://github.com/cartesi/rollups-explorer/actions/runs/12129023863)) It gets stuck in the CI environment. 
   
To check the navigation without the version for the edge case of the same address, different versions 
you can use the following application address on `Sepolia`.

> 0xab7528bb862fb57e8a2bcd567a2e929a0be56a5e
   
PS: I will leave the `NEXT_PUBLIC_PREVIEW_EXPLORER_API_URL` environment variable around as part of the deployment strategy. Later, a PR may come and remove that guy.

PR #285, opened 2025-01-02
#280 Improve balance error for ether deposit form

nevendyulgerov:
I tweaked the validation for the ether deposit amount field so that it's compared to the balance of the logged user. I also covered the new logic with unit tests.

I had to slightly refactor some code related to the unit tests for that form, it's mostly related to the `wagmi` mocks. I had to either continue using the async mocking approach (which we decide to drop) for the new tests or refactor the code so that the standard "synchronous" mocking approach is used throughout the test suite. I went for the latter, I think this is ok because https://github.com/cartesi/rollups-explorer/pull/269 is on hold (which also touches the same files).

### DIFF
--- a/packages/ui/src/EtherDepositForm.tsx
+++ b/packages/ui/src/EtherDepositForm.tsx
@@ -36,6 +36,7 @@ import { useAccount, useWaitForTransactionReceipt } from "wagmi";
 import { TransactionProgress } from "./TransactionProgress";
 import useUndeployedApplication from "./hooks/useUndeployedApplication";
 import { TransactionFormSuccessData } from "./DepositFormTypes";
+import { useFormattedBalance } from "./hooks/useFormattedBalance";
 
 export interface EtherDepositFormProps {
     applications: string[];
@@ -53,6 +54,8 @@ export const EtherDepositForm: FC<EtherDepositFormProps> = (props) => {
     } = props;
     const [advanced, { toggle: toggleAdvanced }] = useDisclosure(false);
     const { chain } = useAccount();
+    const balance = useFormattedBalance();
+
     const form = useForm({
         validateInputOnBlur: true,
         initialValues: {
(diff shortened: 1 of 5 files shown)


brunomenezes:
> I had to slightly refactor some code related to the unit tests for that form, it's mostly related to the wagmi mocks. I had to either continue using the async mocking approach (which we decide to drop) for the new tests or refactor the code so that the standard "synchronous" mocking approach is used throughout the test suite. I went for the latter, I think this is ok because https://github.com/cartesi/rollups-explorer/pull/269 is on hold (which also touches the same files).

That is alright, as it is currently on-hold for a GA or stable release of rollups-v2. Still, from that branch, I will be generating **alpha** releases as necessary to be available to work in conjunction with `cartesi-cli` alphas (cc: @tuler ). So, eventually, I will be doing some rebases there. Also, let's coordinate rebases from that rollups-v2 branch on conflicting changes and you could also open a PR against it later once the changes (e.g. this PR) are merged on **main**, as the changes are fresher on your mind in case it is different from how I refactored the test files based not just on the changes but also the **mocking approach drop** mentioned.

packages/ui/src/EtherDepositForm.tsx
```diff
@@ -53,6 +54,8 @@ export const EtherDepositForm: FC<EtherDepositFormProps> = (props) => {
     } = props;
     const [advanced, { toggle: toggleAdvanced }] = useDisclosure(false);
     const { chain } = useAccount();
+    const balance = useFormattedBalance();
+
     const form = useForm({
         validateInputOnBlur: true,
```
brunomenezes, line 60: @nevendyulgerov, I think validation on input change is better for the current setup. The feedback I get is that when I type a number under my balance, the deposit button is enabled, but at the moment that I enter an invalid or bigger amount, the button disables, but there is no reason.

brunomenezes (changes requested):

brunomenezes (approved):

Issue #287, opened 2025-01-02
Upgrade Next to v15


nevendyulgerov:
## 📄 Context

[Next.js v15](https://nextjs.org/blog/next-15) has been released some time ago so let's test it and upgrade to it.

Please note that there are some breaking changes that will affect the app once we upgrade to `v15`. For example params in server components and `generateMetadata` are now [async](https://nextjs.org/blog/next-15#async-request-apis-breaking-change), while before they were not. So we'll need to handle getting those params using `async/await`.

## 📈 Subtasks

- [x] Upgrade to `v15`.
- [x] Fix all areas with breaking changes.
- [x] Generate a cartesi/rollups-explorer:devel image and use it with your cartesi-cli to check if all available functionalities/pages are working as expected

## Definition of done:

- [x] The app is working as before.
- [x] Unit and e2e tests are passing as before.

brunomenezes:
@nevendyulgerov, please also account for the docker images that are generated and if they are running as expected. For example, generate a `cartesi/rollups-explorer:devel` image and use it with your `cartesi-cli` to check if all available functionalities/pages are working as expected.

PR #289, opened 2025-01-03
#287 Upgrade next to v15

nevendyulgerov:
I upgraded next to `v15` and react to `v19` for the `web` app which resulted in library upgrades in:

- `packages/ui` - upgraded the `react` and `testing-library` versions
- `apps/workshop` - upgraded the `react` and `storybook` versions

I also tested the app through the docker container and didn't find any issues.

Additionally, I disabled all metamask e2e tests as those continue to be flaky. I'll look into those issues in upcoming tasks.


apps/web/src/hooks/useElementVisibility.tsx
```diff
@@ -1,16 +1,12 @@
...
 import { useMemo } from "react";
-type UseElementVisibilityParamsProps<T extends Element> = {
-    element: React.RefObject<T>;
-    threshold?: number;
-};
 
-export const useElementVisibility = <T extends Element>({
-    element,
+export const useElementVisibility = (
+    element: HTMLDivElement | null,
```
brunomenezes, line 5: That is an odd type narrowing saying that the only element type accepted would be `div` elements or `null`. But in theory, it was supposed to accept any valid HTML element or null. I checked the code and played around, and it would take a ref typed as `HTMLInputElement` without any squiggly red lines (the project is using typescript 5.3.2), which causes other problems in terms of modifying that hook as the intellisense will display incorrect attributes.

I believe the element being a generic type that extends the correct `ParentType` is necessary.

brunomenezes: Hey @nevendyulgerov, I am not saying to put it back to what it was (like you're doing here [cbf28cb](https://github.com/cartesi/rollups-explorer/pull/289/commits/cbf28cb301c9b5a445cf5c8be74b76b3c1cb9a63)) as I mentioned refactoring to one level down by receiving the `HtmlElement or null` instead of the `RefObject` is fine, but my highlighting was about the type narrowing to Div, instead of a more generic one _e.g. by using generics._

nevendyulgerov: Hi, @brunomenezes , I'm not sure I understand. The code does use generics as it did before with the changes done in https://github.com/cartesi/rollups-explorer/pull/289/commits/cbf28cb301c9b5a445cf5c8be74b76b3c1cb9a63.

Please let me know if you want further changes related to this.

brunomenezes: To clarify further, the comment was not about the refactoring from the object param to the function param, just the type narrowing.

brunomenezes (approved):

PR #266, opened 2024-11-14
feat: Rollups v2 support

brunomenezes:
### Summary
Code changes to support both Cartesi applications with rollups v1 and v2 (currently available on devnet and sepolia). 

PS: Related to API rollups v2 support https://github.com/cartesi/rollups-explorer-api/pull/52 **(Not deployed)**

PR #290, opened 2025-01-03
Fix flaky e2e test

nevendyulgerov:
Fix for the flaky e2e test that is causing the e2e job to fail.


brunomenezes (approved):
Nice! I will also cherry-pick a few other things from #266

PR #294, opened 2025-01-08
Feat: Ether deposit validation on change and display balance

brunomenezes:
## Summary

Code changes to improve feedback in the ether deposit form. Also, I refactored the test file for the ether-deposit form before adding the new test cases (separated commits). 

Changes: 
- The validation run on input changes.
- The ETH account balance is always visible.
- When the Eth balance is bigger than 0, a `max` button is available to click.
- When the amount has a value entered (i.e. input is considered dirty), and the user changes the account through the wallet, the validation is run again. 
- Testing mock imports refactored for eth-deposit form test file.
- New test cases added.


packages/ui/src/EtherDepositForm/index.tsx
```diff
@@ -54,22 +56,27 @@ export const EtherDepositForm: FC<EtherDepositFormProps> = (props) => {
     } = props;
     const [advanced, { toggle: toggleAdvanced }] = useDisclosure(false);
     const { chain } = useAccount();
-    const balance = useFormattedBalance();
```
nevendyulgerov, line 57: It seems that we won't be needing `useFormattedBalance` hook. If this is true, please delete it and the related unit test suite.

packages/ui/src/EtherDepositForm/index.tsx
```diff
@@ -157,21 +178,44 @@ export const EtherDepositForm: FC<EtherDepositFormProps> = (props) => {
...
+                        rightSectionWidth={60}
+                        rightSection={<Text>ETH</Text>}
+                        withAsterisk
+                        {...form.getInputProps("amount")}
+                    />
+
+                    <Flex c={"dark.2"} gap="3">
+                        <Text fz="xs">Balance: {accountBalance.formatted}</Text>
+                        {accountBalance.value > 0 && (
+                            <UnstyledButton
```
nevendyulgerov, line 199: Can we switch to using the standard `Button` component instead?

nevendyulgerov (commented):
I left a couple of comments.

packages/ui/src/EtherDepositForm/index.tsx
brunomenezes: Removed on [c9c0c7f](https://github.com/cartesi/rollups-explorer/pull/294/commits/c9c0c7f444fbe9f12b997f1294c4a4fd31dd9b0f)

packages/ui/src/EtherDepositForm/index.tsx
brunomenezes: Hey! I started with the `<Button>`, but I needed to unstyle it (no pun intended) to reach the same result, i.e. adding `variant: transparent, height: auto and padding 0` besides the required one for the inner text. Hence, the reason to use the `<UnstyledButton />`.

nevendyulgerov (approved):

Issue #297, opened 2025-01-13
ABI Parameters spec can be created/updated with invalid parameters


nevendyulgerov:
## 🙂 Expected behavior

The ABI Parameters spec form should not accept invalid parameters when creating or updating it.

## 🫠 Actual behavior

The ABI Parameters spec form allow invalid parameters when creating or updating it.

## 🧪 Minimal test case

1. Open the `/specifications/new` page
2. Fill in a name for the spec
3. In the `ABI Parameter` input, insert `invalid abi-spec`
4. Click the `Add` button and notice the error that appears under the field - `Unknown type. Type "invalid abi-spec" is not a valid ABI type. Version: abitype@1.0.7`.
5. Click the `Save` button

After step 5 the spec is created successfully although there's an error for the ABI parameters. Updating the same spec with invalid parameters works just the same.

## 🌎 Environment

Any environment. Any Rollups version.

## ✔️ Possible solutions

The `Add` button (from step 4) should not allow inserting invalid ABI parameters. The validation for this should be triggered at this phase, rather than later when the ABI parameter is already added.

PR #312, opened 2025-01-22
#297 Validate abi params entry when adding it

nevendyulgerov:
I tweaked slightly the logic for adding abi params from the specs form. Now, when clicking on the `Add` button next to the abi params input, we'll first parse the params to ensure they're valid. Only if they are, we'll add them to the entries array.

I kept the validation for entries, this line:
```
const error = isNotNilOrEmpty(entries) ? checkError(entries) : null;
```

so that users can still see an error for the entries if they previously saved a spec with invalid params.

### DIFF
--- a/apps/web/src/components/specification/form/fields/HumanReadableABIParameter.tsx
+++ b/apps/web/src/components/specification/form/fields/HumanReadableABIParameter.tsx
@@ -10,8 +10,8 @@ import {
 } from "@mantine/core";
 import { createFormActions, useForm } from "@mantine/form";
 import { clone } from "ramda";
-import { isFunction, isNotNilOrEmpty } from "ramda-adjunct";
-import { FC, ReactNode, useEffect, useRef } from "react";
+import { isFunction, isNilOrEmpty, isNotNilOrEmpty } from "ramda-adjunct";
+import { FC, ReactNode, useEffect, useMemo, useRef } from "react";
 import { TbTrash } from "react-icons/tb";
 import { parseAbiParameters } from "viem";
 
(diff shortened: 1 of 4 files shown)


brunomenezes (changes requested):
@nevendyulgerov, I think the issue would be more around not allowing to save if it is in a state of error rather than not allowing declarations to be added. As an example, is that with that change I am not allowed to either declare a `Struct` (a.k.a tuple) first or declare an abi parameter signature that uses a tuple ahead of time (i.e. before I declare the tuple) I am limited to inline declarations. My suggestion is to rollback that one and focus on not allowing the save to go through when in an error state, as it does when the name is empty. 

### New changes, I am not able to define a struct.


### Current production deployed. After using a Struct ahead of time, then declaring it.

apps/web/src/components/specification/form/fields/HumanReadableABIParameter.tsx
```diff
@@ -86,7 +89,7 @@ export const HumanReadableABIParameter: FC<HumanReadableABIParameter> = (
                         Add
                     </Button>
                 }
-                error={props.error || form.errors.abiParamEntry}
+                error={isNilOrEmpty(entries) ? props.error : null}
```
brunomenezes, line 92: Hi @nevendyulgerov, this here is doing two different things. The initial is saying that if the caller of that component is setting an error, it has higher precedence than an internal error. However, the change only displays the external set error (if any) if the entries are nil/empty.

nevendyulgerov: Hi, @brunomenezes , looking at the code there's no validation for `abiParamEntry` so the second part of the expression will always evaluate to `null`, so in any case only the external error (coming from `props.error`) can be displayed here.

The reason for this change was that without it, there will be 2 errors for the abi params displayed when users attempt to save an ABI Params spec with invalid params.

Here's a video demonstrating this (using `error={props.error || form.errors.abiParamEntry}`:
https://github.com/user-attachments/assets/01a949d7-1ed8-49fc-be37-cf7c47d202bf

Another video, using `error={isNilOrEmpty(entries) ? props.error : null}`:
https://github.com/user-attachments/assets/79549060-f210-440d-bee5-48a51981bbac

brunomenezes: What about edit mode? I think it is more relevant in that sense.

nevendyulgerov: I think the same logic applies for edit mode.

brunomenezes (approved):

Issue #351, opened 2025-06-16
Add search capability [Applications]


brunomenezes:
## 📄 Context
 It would be a valuable addition to be able to find an application based on its address quickly. This page offers a section specifically for your apps (when connected). However, you may not have an application and still be interested in a specific one or a set of apps owned by a particular address.

## ✔️ Solution
Add an input field to look up for applications based on their `address` and/or `owner`  properties. The query should be integrated with the `query string` part of a URL (e.g. like the input’s search does). 

* Values coming from the search field should be debounced.
* Use fuzzy search to get results. (Could use inputs checkQuery.ts as a reference for query building).
* Values should be added/updated to the query string in the URL.
* The filtered page result should be shareable (e.g copy/paste URL and sending to another user), i.e. the query should accept values from the query string only using pre-defined params, anything else should be ignored by using default values (if any).

**PS: That is a good opportunity to refactor the use of `react-icons/ci` in the input's page search as it is the only place using that set. We’re currently using the `tb` set everywhere.**

## 📈 Subtasks

- [ ] Add the search input field 
- [ ] Integrate the search query into the URL query params.
- [ ] Implement the fuzzy search for Applications. 
- [ ] Add tests.  
- [ ] Update E2E tests.

## 🎯 Definition of Done

- [ ] I can use the search and find a closer enough application by its address. 
- [ ] I can use the search to find the apps based on an owner's address.
- [ ] I can receive a URL shared by another user and see the same results.
- [ ] CI builds are passing. 
- [ ] Preview environment builds are completing.

PR #356, opened 2025-06-23
#351 Add applications search

nevendyulgerov:
I added the applications search that handles application addresses or owners.

Additionally, I refactored the search input removing its wrapper with negative margins and created a slot in the `Paginated` component for it.

Finally, I covered the new logic with unit and e2e tests.

### DIFF
--- a/apps/web/src/components/applications/applications.tsx
+++ b/apps/web/src/components/applications/applications.tsx
@@ -1,6 +1,7 @@
 "use client";
 
 import { Tabs } from "@mantine/core";
+import { useDebouncedValue } from "@mantine/hooks";
 import { FC, useCallback, useState } from "react";
 import { useAccount } from "wagmi";
 import {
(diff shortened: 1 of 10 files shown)


apps/web/src/components/applications/applications.tsx
```diff
@@ -68,18 +71,32 @@ const UserApplications: FC = () => {
...
-    const [query] = useApplicationsConnectionQuery({
-        variables: {
-            orderBy: ApplicationOrderByInput.IdAsc,
-            limit,
-            after,
-            chainId,
+    const formattedQueryDebounced = queryDebounced.toLowerCase();
+    const [{ data: data, fetching: fetching }] = useApplicationsConnectionQuery(
+        {
+            variables: {
```
brunomenezes, line 82: Hi @nevendyulgerov, I suggest extracting that to a function to build the query, which is why I mentioned `query.ts` as a reference, as it is for inputs. There is filtering coming, so it will be easier to extend by passing new parameters to that function without disrupting too much. Additionally, it would be good to replace the `orderBy` from `IdAsc` with `TimestampDesc` so that the newest apps are always displayed at the top.

brunomenezes (changes requested):

Issue #352, opened 2025-06-16
Add filtering capability (by rollups version)


brunomenezes:
## 📄 Context
 As we support multiple Cartesi rollup versions, it would be a good addition to be able to filter based on the application’s version. 

**PS: Just for context, for CartesiScan’s indexer, an application version is defined based on which contract factory it is coming from. Therefore, if it’s coming from `DAppFactory`, it’ll be branded as `v1` and if it is coming from `ApplicationFactory` contract, it will be v2. Previously, we were dealing with inputs submitted to the `InputBox` to applications that do not exist, and we were synthetically creating these apps.  We removed that behaviour since the v1.0.0 release.**

## ✔️ Solution

Add a filter button that expands to a submenu with the rollup version options available. Make room for that component also to accommodate other future filterable properties.

* Add a button with a filter icon aligned with the search input (UI reference attached)
* Display a visual indication of when the filter is in use or not. (UI reference attached)
* Integrate with the query string URL params. As usual, define the parameters accepted and ignore anything else. 
* The user should be able to check multiple versions.
* It is necessary to **apply** the filter changes as a confirmation before it is fired. 
* I can clear the filters. That will uncheck all the boxes and immediately update the URL and query fired. 
* When the menu is open, I can click the filter button again or outside to close it.
* Add filter button to the inputs list page. 
* Add filter button to the applications list page. Depends on https://github.com/cartesi/rollups-explorer/issues/351 UI design-wise.
* **This button should not be rendered or affect the search when listing the inputs of an application. That is because the user is already listing a specific application that has its own version (i.e. under `/applications/:address/:version/inputs`). In that case, that filter is not necessary at all.

## 📈 Subtasks

- [ ] Implement the filter button + menu component. 
- [ ] integrate query params to be part of its state.
- [ ] Add component to the inputs page. 
- [ ] Add component to the applications page. __Depends on another issue, check bullet points above__
- [ ] Update E2E test cases. __Use the bullet points above as a reference for the cases__
- [ ] Add/update existing tests.

## 🎯 Definition of Done

- [ ] In the inputs page, I can filter the inputs related to the rollup version(s) I checked.
- [ ] In the applications page, I can filter the list of applications based on the version(s) I checked.
- [ ] The filter button **is not visible** in any specific application inputs listing page.
- [ ] Filters should only affect the data and URL once I finish selecting and applying it. 
- [ ] I can copy and paste a shared URL with the filters included and obtain the same result.
[...]

PR #359, opened 2025-06-30
#352 Add version filtering

nevendyulgerov:
I added the versions filter to the applications and inputs pages, excluding the pages where the version is predefined (present in the url as a param).

Additionally, I covered the new logic with unit and e2e tests.

### DIFF
--- a/apps/web/e2e/pages/applications.spec.ts
+++ b/apps/web/e2e/pages/applications.spec.ts
@@ -85,7 +85,6 @@ test("should search for specific application", async ({ page }) => {
 test("should filter applications based on rollups v2 version", async ({
     page,
 }) => {
-    await page.reload();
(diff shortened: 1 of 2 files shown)


brunomenezes (changes requested):
Hi @nevendyulgerov, 
Good work, there is one thing to address that I got. The parameter is not checked and is accepting any value. Therefore, the component displays the wrong state, and the list is always empty.

By `checking`, I mean doing something similar to what we do with the `lt` parameter, which is also a known set of values.

URL to reproduce: https://rollups-explorer-sepolia-19jccfz6t-cartesi.vercel.app/applications?query=&pg=1&lt=10&version=v10


Screenshot 2025-07-07 at 07 10 03

brunomenezes (approved):

Issue #370, opened 2025-07-30
Filtering by rollup version doesn't work for multiple versions on page refresh


nevendyulgerov:
## 🙂 Expected behavior

When you've selected both v1 and v2 versions, apply the filters and refresh the page, the versions should be preselected from the url params.

## 🫠 Actual behavior

When you've selected both v1 and v2 versions, apply the filters and refresh the page, the versions are not preselected from the url params.

## 🧪 Minimal test case

1. Go to `/applications`
2. Filter by both v1 and v2 versions
3. Refresh the page

## 🌎 Environment

Mainnet / Sepolia / Local
Any rollups version

PR #371, opened 2025-07-30
#370 Fix issues with filters persistence

nevendyulgerov:
I fixed the issue related to the multiple rollups versions not being correctly persisted on page refresh.

Additionally, I noticed that on production there are cases where 2 requests are being made for the filters. This was caused by the debouncing logic for the filters, so I removed it. Changes to the filters are low-frequency (you need to select the filters and then hit `Apply` or `Reset` buttons after which the filters popover is closed) so there won't be a possibility for the users to spam requests in such ways.

### DIFF
--- a/apps/web/src/components/applications/applications.tsx
+++ b/apps/web/src/components/applications/applications.tsx
@@ -18,6 +18,7 @@ import Paginated from "../paginated";
 import Search from "../search";
 import VersionsFilter from "../versionsFilter";
 import UserApplicationsTable from "./userApplicationsTable";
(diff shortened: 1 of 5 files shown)


brunomenezes:
@nevendyulgerov, You could add a test case for this bug you found to make sure it is covered now, and to avoid future changes reintroducing it.

nevendyulgerov:
> @nevendyulgerov, You could add a test case for this bug you found to make sure it is covered now, and to avoid future changes reintroducing it.

@brunomenezes , I added e2e tests.

brunomenezes (approved):

Issue #378, opened 2025-08-21
Selecting a contract for ERC-20 deposit form results in many requests made to the RPC node provider


nevendyulgerov:
## 🙂 Expected behavior

Selecting a contract for the ERC-20 deposit form shouldn't result in rapid number of requests being made to the RPC node provider.

## 🫠 Actual behavior

Selecting a contract for the ERC-20 deposit form results in rapid number of requests being made to the RPC node provider:

https://github.com/user-attachments/assets/fdbbb6f0-e186-4409-bd6e-f8208fe37804

## 🧪 Minimal test case

1. Connect your wallet
2. Click on the Send Transaction button
3. Select ERC-20 Deposit type
4. Select application
5. Select contract

## 🌎 Environment

All networks
All rollups versions

PR #379, opened 2025-08-21
#378 Fix issue with RPC requests

nevendyulgerov:
The query key returned from `useReadContracts` had unstable reference that caused a re-invocation loop for `useWatchQueryOnBlockChange`. I fixed it by verifying the uniqueness of the query and block number in that hook.

### DIFF
--- a/packages/ui/src/hooks/useWatchQueryOnBlockChange.ts
+++ b/packages/ui/src/hooks/useWatchQueryOnBlockChange.ts
@@ -1,7 +1,7 @@
 import { useEffect, useRef } from "react";
 import { useQueryClient } from "@tanstack/react-query";
 import { useBlockNumber } from "wagmi";
-import { isEqual } from "../utils";
+import { equals } from "ramda";
 
 export default function useWatchQueryOnBlockChange(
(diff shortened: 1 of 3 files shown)


packages/ui/src/hooks/useWatchQueryOnBlockChange.ts
```diff
@@ -1,14 +1,24 @@
...
 
     useEffect(() => {
-        queryClient.invalidateQueries({ queryKey });
+        if (
+            !equals(queryKey, lastQueryKey.current) ||
+            blockNumber !== lastBlockNumber.current
+        ) {
+            lastQueryKey.current = queryKey;
+            lastBlockNumber.current = blockNumber;
+            void queryClient.invalidateQueries({ queryKey });
```
brunomenezes, line 21: @nevendyulgerov, why the void signalling?

nevendyulgerov: I'm using `void` here because we are ignoring the returned promise from `invalidateQueries`.

brunomenezes (approved):

PR #401, opened 2025-12-01
Node API

tuler:
This PR introduces the usage of the rollups-node updated JSON-RPC API, that also contains tournament information.
The node has not released a version yet, so this matches a "hot" unmerged branch `feature/pr-draft-reader-mode`.
The API is wrapped by the `@cartesi/viem` and `@cartesi/wagmi` libraries, which have alpha releases (still prone to changes).

The types defined in the project have been replaced, for the most part, by types defined at `@cartesi/viem`, which match the API information.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -57,7 +57,7 @@
     "devDependencies": {
         "@cartesi/tsconfig": "workspace:*",
         "@playwright/test": "^1.51.0",
-        "@tenkeylabs/dappwright": "^2.11.2",
+        "@tenkeylabs/dappwright": "^2.11.4",
         "@testing-library/jest-dom": "^6.6.3",
         "@testing-library/react": "^16.3.0",
         "@testing-library/user-event": "^14.6.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -305,8 +305,8 @@ importers:
         specifier: ^1.51.0
         version: 1.54.1
       '@tenkeylabs/dappwright':
-        specifier: ^2.11.2
-        version: 2.11.2(playwright-core@1.54.1)
+        specifier: ^2.11.4
+        version: 2.12.0(playwright-core@1.54.1)
       '@testing-library/jest-dom':
         specifier: ^6.6.3
         version: 6.6.3
@@ -3083,8 +3083,8 @@ packages:
     peerDependencies:
       react: ^18 || ^19
 
-  '@tenkeylabs/dappwright@2.11.2':
-    resolution: {integrity: sha512-ITIY1yKqSMoupuu3kKvUMdXOWUg4+T5O5NZLQKIuAf84zk7oxoIjIAgV2Go5QGynQTpSZhKK7o7YmE93PKPWtQ==}
+  '@tenkeylabs/dappwright@2.12.0':
+    resolution: {integrity: sha512-AVGT827BFjNWqz6epsXeAy6Tref8K0iviuogIrAdokPXaVj/8te6Pf+F9dKF0Xy6eE3oXEbqD5+SCZyvb0FQSQ==}
     engines: {node: '>=20'}
     peerDependencies:
       playwright-core: '>1.0'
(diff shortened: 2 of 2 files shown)


tuler:
The rationale of this approach is to use the models as defined by the JSON-RPC API. (the viem/wagmi are only typescript adaptations and hooks for the API, no additional logic is added). The main goal is to reduce boilerplate code and facilitate maintenance, at the cost of bringing the view closer to the model of the API, which in some cases can make it difficult to evolve with future versions (which I'm not worried about).

There are two ways to "inject" mock data.

1) create alternative containers that use mock data directly, instead of fetching data. There is already a separation of view components, which are presentation only, and container components, which fetch data.

2) use `queryClient.setQueryData` to inject mock data into the query client, which will deliver the data instead of fetching.

apps/dave/src/components/epoch/EpochCard.tsx
```diff
@@ -1,26 +1,33 @@
...
+    const params = useParams<EpochParams>();
     const epochIndex = epoch.index?.toString() ?? "0";
-    const url = routePathBuilder.appEpochDetails({ ...params, epochIndex });
+    const url = routePathBuilder.epoch({
+        application: params.application ?? "",
+        epochIndex,
+    });
     const isMobile = useMediaQuery(`(max-width: ${theme.breakpoints.sm})`);
     const color = useEpochStatusColor(epoch);
+    const inDispute = false; // XXX: how to know if an epoch is in dispute?
```
brunomenezes: @tuler, just a comment/discussion; In the current state of the node-api, I believe the app would need to check for tournament and matches (if there is no matches it is not really in a "dispute" state). Also, that "search" would only make sense in a closed status epoch, because I understand that a `Finalized` one means the tournament is also finished, right?

apps/dave/src/containers/TournamentContainer.tsx
```diff
@@ -1,50 +1,80 @@
...
-
     return (
         <Stack pt="lg" gap="lg">
             <Hierarchy hierarchyConfig={hierarchyConfig} />
 
             {isLoading ? (
                 <ContainerSkeleton />
-            ) : tournament !== null ? (
-                <TournamentPage tournament={tournament} />
+            ) : !!tournament ? (
```
brunomenezes: just 2cents. I think the use of `isNotNil` from ramda is clearer than double negation.

apps/dave/src/containers/EpochContainer.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    return (
+        <Stack pt="lg" gap="lg">
+            <Hierarchy hierarchyConfig={hierarchyConfig} />
+
+            {isLoading && <ContainerSkeleton />}
+            {!!epoch && (
+                <EpochDetailsPage epoch={epoch} inputs={inputs?.data ?? []} />
+            )}
+            {!isLoading && !epoch && (
```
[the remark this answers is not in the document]
tuler: the style of ternary statements in JSX sometimes become very weird IMO.

apps/dave/src/containers/TournamentContainer.tsx
brunomenezes: Ramda is a functional library toolkit if you will. You can say it is also a "matter of style". For that specific function you would be replacing `tournament !== null && tournament !== undefined` by it.

apps/dave/src/providers/DataProvider.tsx
```diff
@@ -6,7 +7,9 @@ const queryClient = new QueryClient();
 const DataProvider: FC<PropsWithChildren> = ({ children }) => {
     return (
         <QueryClientProvider client={queryClient}>
-            {children}
+            <CartesiProvider rpcUrl="http://127.0.0.1:10011/rpc">
```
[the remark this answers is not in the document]
brunomenezes: > This is definitely temporary, it will be replaced by a UI soon.

By an UI?

PR #418, opened 2026-01-23
feat: ERC-20 deposit

### LINKED ISSUE #415: Portals: Implement ERC-20 deposit
## 📄 Context

An application accepts ERC-20 deposits via interactions with the ERC20Portal (Rollups version 2). Also, before the deposit can be done, an allowance must be set to the ERC20Portal (a.k.a Spender)

depends on: #413 

A consideration is that submitting a transaction should not work or be reflected when the `mock` data is enabled.

The Dave webapp is node-API-centric, and the UI currently provides a top-down view from the application perspective. Therefore, a `Send` select will be added to the existing application card, so the form, by default, always has the `application address` (Common to all upcoming Portals and InputBox transactions)

Implement the form to send an ERC-20 token with/without execution data to a specific application. 

Requirements:
* Can only make the deposit if a wallet is connected.
* Can only click the deposit button if there are funds.
* Can only click the deposit button if the amount is bigger than zero.
* Can only click the deposit button if the amount is equal or smaller than the allowance. 
* Should be able to set an allowance to the spender (i.e. ERC20Portal address)
* Should display any transaction errors in the form. 
* Upon transaction confirmation, notify of success with a `chain explorer` link when such a chain is not a devnet.

Deployment Reference about ERC-20 Deposit: [ERC20Portal](https://usecannon.com/packages/cartesi-rollups/2.1.0/1-main/interact/cartesi-rollups/ERC20Portal/0xACA6586A0Cf05bD831f2501E7B4aea550dA6562D)
Documentation(`staging`) about ERC-20 Portal: [ERC20Portal docs](https://staging.mugen.builders/cartesi-rollups/2.0/api-reference/contracts/portals/ERC20Portal/)

## 📈 Subtasks

- [x] Implement the deposit form.
- [x]  Implement the `depositERC20Tokens` transaction lifecycle.

## 🎯 Definition of Done

- [x] I can deposit any ERC-20 token with success to a specific Cartesi Application.

brunomenezes:
# Summary
Code changes to add support for ERC-20 deposits. A big chunk of code was ported from the current rollups-explorer, with refactorings to target the data structures returned by the rollups-node API and to use cartesi/viem and cartesi/wagmi.

A test can be performed using the latest Cartesi CLI, ensuring the mock is disabled (default). Also, a devnet deploy of a simple ERC-20 token is necessary.

### DIFF
--- a/apps/dave/src/app/layout.tsx
+++ b/apps/dave/src/app/layout.tsx
@@ -12,6 +12,24 @@ interface RootLayoutProps {
     children: ReactNode;
 }
 
+/**
+ * This is a workaround to solve a problem when react-dom-client.development
+ * is trying to stringify bigints.
+ *
+ * refer to issue {@link https://github.com/facebook/react/issues/35004}
+ * a PR is currently open here {@link https://github.com/facebook/react/pull/35013}
+ */
+if (process.env.NODE_ENV === "development") {
+    Object.defineProperty(BigInt.prototype, "toJSON", {
+        writable: false,
+        enumerable: true,
+        configurable: false,
+        value() {
+            return this.toString();
+        },
+    });
+}
+
 const RootLayout: FC<RootLayoutProps> = ({ children }) => {
     return (
         <html lang="en" {...mantineHtmlProps}>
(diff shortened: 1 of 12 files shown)


tuler (commented):
Good that we are able to reuse code.
Are these hooks the ones generated by Wagmi CLI, or manually written?

tuler (dismissed):

PR #449, opened 2026-04-02
feat: add outputs list page

### LINKED ISSUE #428: Output: Add a list page
## 📄 Context

The rollups-node API can provide a list of all outputs for an application. 

It will benefit from predecessor issues as listed below for UI elements reuse.
**Depends on**: #426 , #427 

## ✔️ Solution

Display a list of the existing outputs for a given application.

* Paginated amount of outputs.
* Filterable by types(i.e. `Notice`, `Voucher`, `DelegateCallVoucher`)
* Can execute Outputs of type `Voucher` and `DelegateCallVoucher` when not executed already.

## 📈 Subtasks

- [ ] Implement the filterable list of outputs.

## 🎯 Definition of Done

- [ ] I can see a paginated list of Outputs
- [ ] I can execute `Voucher` and `DelegateCallVoucher` output types.

brunomenezes:
# Summary
Code change to add a page to access outputs of a specific application from its summary page (small recordings below). 

What's new: 
* Added outputs page to list outputs from a specific application
* Users can, besides listing everything, also narrow the search by output types (.i.e. Voucher, Delegate call voucher, notice)
* The search can be sorted (default: Descending)
* The search result has a limit of items per page (i.e current implementation values are 10 / 30 / 50). **Default: 50**
* The search primarily respects the use of the URI query component (more info here on [RFC 2396 section 3.4](https://datatracker.ietf.org/doc/html/rfc2396#section-3.4)). The query parameters are sanitised, and if not matched, defaults are used instead, except for the `offset`, which only makes sure it is a number and not negative 
* As the query component is used, sharing the current URL should display the same search result.
* Auto-redirect to the home when the node-connection is updated was removed. **Affected containers handle the possible 404s if that is the case.**  

**PS: Further search narrowing is out of scope for this implementation, like the use of the `voucher_address` property.**

## Short clips

<details>
<summary>Access the output's page</summary>

</details>

<details>
<summary>Narrowing the search by output type</summary>

</details>

<details>
<summary>Direct access to the Epoch's details</summary>

</details>


guidanoli (approved):
Tested locally with a dummy application and looks good to me!

